PR #12, opened 2026-05-06
feat(dsl): detail-modeling phases A–C + organic primitives & stdlib wrappers

oldschoola:
> **Update 2026-05-07 (commit `36255a3`):** Z.ai (GLM) chat surface fixes — drop unsupported `seed` field and round-trip `temperature` through Display→f64 to avoid the f32-widening artifact (both 400'd live requests). Vision input wired up: `glm-5v-turbo` auto-swap when an image is attached, OpenAI-compatible `image_url` content shape, `Provider::supports_images()` now returns true for Z.ai (so `--auto-refine` works against GLM). New Studio Plan checkbox in New from Prompt (`Settings::plan_first`) wires `generate_plan` + `compose_coder_prompt` into `run_llm`. Refine panel adds a Z.ai-only "Use GLM-5V-Turbo for refine" toggle. Two demo `.mog` examples committed (`zai_text`, `zai_plan`); see `examples/README_zai.md` for the two that hit Z.ai connection drops on heavy SI calls.

---

# Detail-modeling: more shapes, smarter LLM, friendlier authoring

12 commits, 71 files, +6960 / -208. Each commit is independently buildable and tested. The summary below skips the per-line reasoning (the commit messages have it) and just explains, in plain English, what's new and why a reviewer should care.

---

### New primitives

| New | What it makes |
|---|---|
| `extrude` | Push a 2D shape straight up — I-beams, gear teeth, picture frames |
| `sweep` | Drag a 2D shape along a curve — pipes with non-circular cross-sections |
| `loft` | Stack different cross-sections at different heights — boat hulls, bottles |
| `chamfered_box` | A box with flat 45° bevels (the sharp-edge cousin of `rounded_box`) |
| `inset_box` | A box with one face pushed inward — control panels, switch wells |
| `coil` | A helix tube — springs, coiled rope, wire wraps |
| `wave` | A deformer that ripples a surface — water, banners, flags |
| `heightfield` | An fbm-noise grid — terrain |
| `bezier_patch` | A smooth bicubic surface from 16 control points — fenders, body panels |
| `metaball` | Smooth-blended spheres — blobs, soft creatures |

### New stdlib modules (call with `use "name" (...)`)

**Detailing motifs** — `bolt_circle`, `vent_strip`, `panel_seam`, `rivet_line`, `step_taper`, `cable`, `chain`, `feather_card`, `scale_band`, `gear`. The LLM can reach for hard-surface decoration in one call instead of placing dozens of primitives.

**Organic wrappers** — `spring`, `terrain_patch`, `blob`, `water_patch`. Sensible defaults over the new primitives so the LLM doesn't have to retune fbm octaves or metaball blend factors from scratch.

### New language features

- **`if` / `else`** — emit children only when a condition is true. Conditions use the new comparison operators (`<`, `<=`, `>`, `>=`, `==`, `!=`).
- **`for (var=…, from=…, to=…[, step=…])`** — emit the body N times, binding the loop variable. Capped at 100k iterations so a typo can't OOM.
- **String interpolation** — `"leg_$i"` or `"leg_${i}"` inside string attributes and node names. Integers render without `.0`.
- **Per-node `lod=N`** — multiply the active LOD scale for one subtree (e.g. `lod=2` for a hero part, `lod=0.5` for background filler).
- **Per-modifier `*_range=[a, b]`** — gate any deformer (`bend_x/y/z`, `twist_y`, `taper`, `droop`, `noise`, `jitter`) to a normalised slice along the kernel's length. Bend the tip of a sword while the grip stays straight; roughen only the crown of a cliff. Backwards-compatible — no flag passed = full effect everywhere, same as before.

### Bug fix: BOM-prefixed files

Some Windows editors silently insert a UTF-8 byte-order mark at the start of saved files. The grammar now tolerates it (in the `file` rule, so diagnostic spans still point at the right byte).

---

### Architect / Coder split (`--plan`)

When you pass `--plan`, generation runs in **two LLM calls** instead of one:

1. **Architect** — text-only call that returns a Markdown plan (subject, parts, hierarchy, materials). Cheap, no DSL emitted.
2. **Coder** — gets the original prompt **plus** the plan, and writes the actual DSL.

Why split? The model is best at spatial reasoning in plain language. Asking it to plan and emit syntax in one shot is what produces the "drowning in primitive coordinates" failure mode.

Defenses on the planner output:
- If the planner ignores its instructions and emits DSL keywords, we warn on stderr and strip Markdown fenced blocks before the plan reaches the Coder. Stops phantom DSL from priming the second call.

### Visual self-critique (`--auto-refine N`)

After a successful generate, the CLI can render the DSL with the headless renderer, send the PNG back to a vision-capable LLM with the original prompt, and let the model rewrite its own DSL based on what it actually sees. Repeats `N` times. Each iteration runs through the full validate+repair loop so the final file still compiles.

- **Capped at 0..=10.** A typo (`--auto-refine 100`) can't burn through hundreds of vision calls.
- **Reviewer LLM failures past iteration 1 are recoverable.** A 503 on call 2 doesn't throw away the perfectly-good DSL produced by call 1 — the spinner abandons with a warning and the loop keeps the latest valid outcome. (Render failures already worked this way; the asymmetry is now gone.)
- **In `modify` mode**, the Reviewer's "original prompt" is `{file's meta(prompt=…)}\n\nMost recent edit applied: {edit instruction}`, so the model critiques against the **asset description**, not the verb-phrase edit (`"make the legs taller"`).

### Gemini auth UX

`--provider {auto,gemini,gemini-oauth,antigravity}` collapses the credential surface into a single `--auth` selector under `Provider::Gemini`. `auto` (the new default) tries gemini-cli OAuth, then Antigravity OAuth, then `GEMINI_API_KEY`. Help text on `--provider` now points at `mogen auth login` instead of listing env vars.

> **Behaviour change:** `--provider` default flipped from `gemini` to `auto`. Scripts that depended on the API-key-only path should pass `--provider gemini` explicitly.

---

## What was caught in three reviewer passes

This branch went through three structured review passes — one of 16 lanes on the geometry/grammar work, one of 19 lanes on the LLM pipeline work, and a final 16-lane sweep across the full diff. All P1 and P2 findings in our own commits landed before merge.

**First pass (geometry / grammar) — `e9ab35d`:**
- `extrude` side-rib winding was inverted; every extruded mesh shipped inward-facing normals. Fixed.
- `sweep` start/end cap winding flags were swapped; both caps faced into the solid. Fixed.
- `for` loop was vulnerable to runaway input — added 100k iteration cap with a friendly error.
- BOM tolerance moved from `parse()` (which shifted every diagnostic span by 3 bytes) into the grammar.
- Several missing test paths and one mis-named test fixed.

**Second pass (LLM pipeline) — `1b54ce0`:**
- `--auto-refine` cap.
- `modify` Reviewer prompt fix.
- Planner DSL-leak guard.
- Reviewer LLM failure asymmetry.
- `visual_refine` config rebuilt around `GenerateConfig::new` so future `GenerateConfig` fields can't silently leak Coder state into the Reviewer call.
- `mogen_dsl::stdlib_registry()` hoisted out of the refine loop.
- Two new mock-server integration tests pinning the planner→coder handoff and the reviewer's image+DSL wiring end-to-end.
- One new lower-stage error test in `refine_render`.

**Third pass (full-diff sweep, P1+P2) — `899548d`:**
- `loft` side-quad winding inverted (`[a,b,d, a,d,c]` → `[a,d,b, a,c,d]`). Every lofted mesh was rendering inside-out. The first pass caught the same bug in `extrude`/`sweep` but missed loft. Fix verified by cross-product on a unit-square loft. New regression test `loft_side_quads_face_outward` locks each cardinal face's normal direction so this can't regress silently.
- `loft` cap winding flags swapped — earcut's CCW 2D triangles map to `-Y` when embedded as `[p[0], h, p[1]]`, so the bottom cap should not flip and the top cap should. New regression test `loft_caps_face_outward` averages position `Y` to disambiguate per cap.
- `interpolate_string` corrupted multi-byte UTF-8. The byte-by-byte loop did `out.push(b as char)`, which mapped each continuation byte to its `U+0080..U+00FF` codepoint — so `"pièce_$i"` rendered as `"piÃ¨ce_0"`. Fix: track `lit_start` and flush literal runs via `&s[lit_start..i]` string slicing. New test covers both `$name` and `${name}` paths with accented and CJK text.
- `inset_box` ignored `UvMode` — every inset quad emitted Fit unit-square UVs regardless of mode, creating a UV scale seam at the inset boundary in Tile mode. Now emits world-space `[p[perp1], p[perp2]]` in Tile so the inset ring tiles at the same metric scale as the surrounding box face. Unused `fit_uv` parameter dropped.
- `loft` silently produced degenerate mesh on duplicate heights. Now bails up front with `"loft sections must have distinct heights"`.
- `chain.mog` claimed "interlocked" but the geometry produced side-by-side links with a 5mm gap (link_b at `x=2*link_radius+gap` is too far for its tube ring to thread link_a's hole — and with equal major radii the topological Hopf link is geometrically impossible without overlap). Tightened link_b shift to `link_radius` so the rings visually overlap by one wire diameter, and updated the summary to be honest about decorative (non-topological) intent.
- `rivet_line.mog` produced NaN geometry on `count=1` because `step=length/(count-1)` divided by zero, and grid's `center_offset` evaluated `0 * inf = NaN`. Now branches on `if (cond=$count > 1)` (using the Phase C3 control flow added earlier in this branch) with an `else` arm that emits a single rivet.

**Out of scope of this PR (filed separately against upstream):**
- `mogen moghub publish` has no client-side size cap on texture uploads (introduced by upstream `50ef4bb Add mogen moghub CLI surface`).
- `mogen://` URL handler in Studio has a TOCTOU race between `version_detail` and `download_zip` on pinned URLs (introduced by upstream `5b7a849 Open MoGHub models from a mogen:// URL in Studio`).
---

## Verification (whole branch, post-rebase)

| Suite | Result |
|---|---|
| `cargo check --workspace` | clean |
| `cargo test -p mogen-geom --features csg` | **96** passed (3 new loft regression tests) |
| `cargo build -p mogen-geom --no-default-features` | clean |
| `cargo test -p mogen-dsl --lib` | **207** passed (2 new interpolation/duplicate-heights tests) |
| `cargo test -p mogen-llm` (lib + 4 integration suites) | 238 + 7 + 8 + 3 + 4 passed |
| `cargo test -p mogen-validate` | 45 passed |
| `cargo test -p mogen --test goldens` | 17 passed (incl. 5 new organic-shape goldens) |
| `cargo test -p mogen --bin mogen refine_render` | 2 passed, 1 ignored (winit + main-thread) |
| `cargo check -p mogen-wasm` | clean |
| `mogen build examples/{cup,sword}.mog` | exit 0 |
| `mogen generate "x" --auto-refine 11 --dry-run` | clap range error before any LLM call |

`crates/mogen/tests/fixtures/mirror_test.glb` was regenerated against post-rebase codegen (1-byte float-arithmetic-ordering drift after `1862a74 Bake mirror reflections into geometry instead of negative scale`). Verified deterministic — re-running goldens produces no further diff.

---

## Out of scope (deliberately)

- **True half-edge mesh chamfer/fillet** on arbitrary input meshes. Would need a half-edge layer `mogen-geom` doesn't have. Primitive-form `chamfered_box` covers ~90% of authoring intent.
- **Animation hook for `wave_phase`** (currently static). Worth wiring up only when an animated water surface becomes a real ask.
- **Marching-cubes / SDF metaball.** Vertex-fillet `union_smooth` handles typical organic blends; see comment in `crates/mogen-geom/src/csg_smooth.rs`.
- **Reviewer system-instruction caching.** Currently sends the ~40 KB Reviewer preamble inline on every refine iteration. Tracked as a follow-up.
- **Planner sees the existing file (in `modify`).** For small edits, blind planning is fine; structural edits ("add a roof") would benefit from a top-level summary fed to the planner. Tracked as a follow-up.

---

## Commit list (13, post-rebase on `master`)

1. `0346e3c` feat(dsl): Phase A — detailing stdlib + per-node lod override
2. `16ccf1e` feat(dsl): Phase B — extrude, sweep, loft primitives
3. `d8ca15d` feat(dsl): Phase C1 — per-modifier `*_range` selective deformation
4. `8b33fab` feat(dsl): Phase C2 — chamfered_box and inset_box primitives
5. `456428c` feat(dsl): Phase C3 — if/else/for control flow + string interpolation
6. `e4daba4` fix(dsl): strip leading UTF-8 BOM in parse()
7. `26faf92` feat(dsl): organic shape primitives — coil, wave, heightfield, bezier_patch, metaball
8. `89001fa` feat(dsl): organic shape stdlib wrappers — spring, terrain_patch, blob, water_patch
9. `299db04` chore(validate): drop unreachable extrude/taper schema arm
10. `e9ab35d` fix(dsl,geom,llm): address PR #12 review findings
11. `1b54ce0` feat(llm): architect planner + visual reviewer refinement loop
12. `ce9577c` chore(goldens): regenerate mirror_test fixture after rebase on master
13. `899548d` fix(geom,dsl): address second-pass review findings


krazyjakee (changes requested):
Awesome additions! High level concern that the LLM may be drowned by the primitives so we may need some kind of intelligent steering. Need to think about it.

Needs to be rebased but then it's good to merge.

krazyjakee (approved):

oldschoola:
> Awesome additions! High level concern that the LLM may be drowned by the primitives so we may need some kind of intelligent steering. Need to think about it.
> 
> Needs to be rebased but then it's good to merge.

I agree, good idea,  https://github.com/threedle/ll3m 
I like how this has a auto refinement, i might add something to mogen like that.

krazyjakee:
Thanks so much for this work, @oldschoola — really appreciate it. Your commits have been rebased onto current `master` and merged forward as part of [#21](https://github.com/krazyjakee/MoGen/pull/21). Closing this one as superseded — the credit and authorship of every commit is preserved in the rebased branch.

PR #13, opened 2026-05-06
Add binary FBX 7.4 export alongside GLB

oldschoola:
## Summary

Adds binary FBX 7.4 export as a sibling to the existing GLB pipeline, behind a new opt-in `fbx` Cargo feature on `mogen-export`. The CLI's `mogen build` learns to dispatch on the output extension (`.fbx` selects FBX) or an explicit `--format glb|fbx` flag.

ASCII FBX, FBX import, and any other format are out of scope.

## Public API

In `mogen-export`, all gated `#[cfg(feature = "fbx")]`, mirroring the GLB shape:

```rust
pub fn write_fbx(scene: &SceneGraph, out: &Path) -> Result<()>;
pub fn write_fbx_with_options<F: Fn(&str)>(scene, out, opts, progress) -> Result<()>;
pub fn build_fbx_with_options<F: Fn(&str)>(scene, opts, progress) -> Result<Vec<u8>>;
pub fn build_fbx_with_options_and_source<F: Fn(&str)>(scene, opts, source, progress) -> Result<Vec<u8>>;
```

`ExportOptions`, `TextureSource`, and the `merge` feature interaction are reused unchanged — the merge passes run before format-specific emission, same as for GLB.

## Crate choice

Picked `fbxcel = "0.9"` with `features = ["writer", "tree"]`. Pure Rust, Apache-2.0/MIT, no C/C++ dependency. The high-level `Tree` API is a perfect fit: emitters build subtrees in-memory then `Writer::write_tree` serialises them, which keeps Definitions/Objects/Connections in sync without a two-pass walk.

Rejected alternatives: `fbxcel-dom` (incomplete), `fbx_direct` (deprecated → fbxcel), `fbx` (reader-only), SmallFBX (C++).

## What's emitted

| Source | Target | Notes |
|---|---|---|
| `SceneNode` | `Model` (Mesh/Light/Null) | TRS via `Lcl Translation/Rotation/Scaling`; quat → Euler XYZ degrees with `RotationOrder=0` |
| `Mesh.positions/normals/uvs` | `Geometry/Vertices` + `LayerElementNormal`/`LayerElementUV` | ByVertice direct mapping |
| `Mesh.indices` | `PolygonVertexIndex` | Last index of each triangle negate-and-decremented (`!idx`) per FBX spec |
| `Material` (PBR) | `Material` (Phong) | DiffuseColor/EmissiveColor/ShininessExponent for the Phong path; raw `Roughness`/`Metallic`/`Transmission`/etc. as `Properties70` entries so PBR importers recover them losslessly |
| `TextureRef` | `Texture` + `Video` | Bytes embedded raw via `Video.Content`; OP-connected to material on `DiffuseColor`/`NormalMap`/`EmissiveColor`/`AmbientColor`/`MetallicRoughnessTexture` |
| `Skin` | `Deformer` (Skin) + per-joint `Deformer` (Cluster) | `Transform`/`TransformLink` from the inverse-bind matrices |
| `Clip` | `AnimStack` + `AnimLayer` + `AnimCurveNode` + 3× `AnimCurve` | KTime ticks (46_186_158_000/sec); rotation tracks quat → Euler XYZ at each keyframe |
| `Light` | `NodeAttribute` (Light) | LightType 0/1/2 (Point/Directional/Spot), FarAttenuationEnd from `range`, InnerAngle/OuterAngle (deg) for spot |
| DSL extras (`role`, `tags`, `cast_shadow`, `kind`) | `Properties70` custom props | Preserved for round-trip tooling |

Coordinate system: `GlobalSettings` advertises Y-up, right-handed, -Z forward, meters (`UpAxis=1`, `UnitScaleFactor=1.0`) — matches glTF and the codebase's de-facto convention. Blender 4.x respects this.

## CLI

```sh
mogen build foo.mog --out foo.glb            # GLB (unchanged default)
mogen build foo.mog --out foo.fbx            # FBX via extension sniffing
mogen build foo.mog --format fbx             # FBX with default-extension override → foo.fbx
mogen build foo.mog --format glb --out x.bin # forced container, ignoring the extension
```

`docs/cli.md` is updated.

## Test plan

New `crates/mogen-export/tests/fbx.rs`, modelled on `tests/lights.rs`. Uses `fbxcel::tree::any::AnyTree` for read-back assertions.

- `top_level_structure_matches_fbx_7_4_layout` — magic header + 7 required top-level sections (`FBXHeaderExtension`, `GlobalSettings`, `Documents`, `Definitions`, `Objects`, `Connections`, `Takes`).
- `global_settings_carry_y_up_and_meter_units` — `UpAxis=1`, `UnitScaleFactor=1.0`.
- `mesh_round_trip_emits_geometry_with_negate_terminated_polygons` — `Vertices` length divisible by 3, every third `PolygonVertexIndex` entry is negative, OO connections Geometry → Model → RootNode.
- `material_with_base_color_texture_emits_full_chain` — Phong shading, `Roughness`/`Metallic` in Properties70, exactly one `Texture` + one `Video`, embedded `Content` bytes match the input PNG, OP `Texture → Material` on `DiffuseColor`.
- `directional_point_and_spot_lights_emit_node_attributes` — three `NodeAttribute` objects, light types 0/1/2, range → FarAttenuationEnd, InnerAngle/OuterAngle in degrees.
- `skin_emits_one_skin_deformer_with_per_joint_clusters` — one Skin Deformer, two Cluster Deformers, each with non-empty Indexes/Weights and 16-element Transform/TransformLink.
- `translation_animation_emits_stack_layer_curve_node_and_three_axes` — one `AnimStack`/`AnimLayer`/`AnimCurveNode`, three `AnimCurve` (X/Y/Z), 3 KeyTime + 3 KeyValueFloat each, last tick within ±1 of `1.0s × 46_186_158_000`, OP CurveNode → Model on `Lcl Translation`.

## Verification

| Check | Result |
|---|---|
| `cargo build -p mogen-export` (default features, no fbx) | ✓ |
| `cargo build -p mogen-export --features fbx` | ✓ |
| `cargo build -p mogen` (transitively pulls fbx) | ✓ |
| `cargo test -p mogen-export --features fbx` | ✓ 39/39 (lib 18 + fbx 7 + lights 4 + textures 10) |
| `cargo test -p mogen --test goldens` | unchanged from baseline (3 pass / 9 pre-existing failures unrelated) |
| End-to-end `mogen build examples/wooden_crate_simple.mog -o /tmp/crate.fbx` | ✓ 76.7 KiB FBX with valid magic + 31 Models / 17 Geometries / 2 Materials / 65 connections |
| End-to-end `mogen build examples/windmill.mog -o /tmp/windmill.fbx` (animated) | ✓ 382.8 KiB FBX with full Animation* + Texture/Video chain |
| `--format` overrides extension sniffing | ✓ verified both directions |
| `docs/cli.md` updated | ✓ |

The independent JS-based FBX 7.4 binary walker (separate code path from the `fbxcel`-based reader used in tests) confirms structural validity end-to-end — top-level sections, Object-type counts, animation curves, and embedded texture content all parse cleanly.

## Files

**Created:** `crates/mogen-export/src/fbx/{mod,doc,ids,nodes,mesh,material,texture,skin,anim,light}.rs`, `crates/mogen-export/tests/fbx.rs`.

**Edited (minimal):** `crates/mogen-export/{Cargo.toml,src/lib.rs}`, `crates/mogen/{Cargo.toml,src/main.rs,src/commands/build.rs}`, the existing build callers in `commands/{animate,generate,modify,repair,textures}.rs` (one-arg threading), `docs/cli.md`, `Cargo.lock` (dependency add).

**Untouched:** `mogen-core`, `mogen-geom`, `mogen-render`, `mogen-validate`, `mogen-studio`, `mogen-wasm`, `mogen-llm`, every existing `mogen-export` source file other than `lib.rs` + `Cargo.toml`, every existing test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,6 +34,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "adler32"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+
+[[package]]
+name = "ahash"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
+dependencies = [
+ "getrandom 0.2.17",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1035,7 +1052,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8ac2645a9bf4826eb4e91488b1f17b8eaddeef09396706b2f14066461338e24f"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "bytemuck",
  "document-features",
  "egui",
@@ -1070,7 +1087,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "53eafabcce0cb2325a59a98736efe0bf060585b437763f8c476957fb274bb974"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "emath",
  "epaint",
  "log",
@@ -1083,7 +1100,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d00fd5d06d8405397e64a928fa0ef3934b3c30273ea7603e3dc4627b1f7a1a82"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "bytemuck",
  "document-features",
  "egui",
@@ -1102,7 +1119,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0a9c430f4f816340e8e8c1b20eec274186b1be6bc4c7dfc467ed50d57abc36c6"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "arboard",
  "egui",
  "log",
@@ -1119,7 +1136,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0e39bccc683cd43adab530d8f21a13eb91e80de10bcc38c3f1c16601b6f62b26"
 dependencies = [
- "ahash",
+ "ahash 0.8.12",
  "bytemuck",
  "egui",
  "glow 0.14.2",
@@ -1185,7 +1202,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a32af8da821bd4f43f2c137e295459ee2e1661d87ca8779dfa0eaf45d870e20f"
 dependencies = [
  "ab_glyph",
- "ahash",
+ "ahash 0.8.12",
  "bytemuck",
  "ecolor",
  "emath",
@@ -1249,6 +1266,18 @@ name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
+
+[[package]]
+name = "fbxcel"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1f9e966e38cb462819635792e0a9ee1d6cc5fff9bcdbf7baefb350221a7ae3df"
+dependencies = [
+ "indextree",
+ "libflate",
+ "log",
+ "string-interner",
+]
 
 [[package]]
 name = "fdeflate"
@@ -1689,6 +1718,15 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
+version = "0.11.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ab5ef0d4909ef3724cc8cce6ccc8572c5c817592e9285f5464f8e86f8bd3726e"
+dependencies = [
+ "ahash 0.7.8",
+]
+
+[[package]]
+name = "hashbrown"
 version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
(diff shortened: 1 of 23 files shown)


oldschoola:
Before flagging this for review I ran an 8-agent self-review pass over the diff. Findings and fixes (all in `7db96cd`):

## P0 (1)

- **`skin.rs` — Cluster→joint OO direction was reversed.** The Cluster must be the source/child and the joint `Model` the destination. With the wrong direction, importers would have failed to associate Clusters with their bones and skinning would have been broken end-to-end. Fixed: `connect_oo(cluster_id, joint_model_id)`.

## P1 (7)

- **`anim.rs` — `KEY_ATTR_FLAGS_LINEAR` (0x4) and `KEY_ATTR_FLAGS_CONSTANT` (0x2) were swapped.** Cross-checked against the FBX SDK `kfcurve.h` and Blender's `io_scene_fbx` exporter. Linear tracks were exporting as step keyframes and vice versa.
- **`texture.rs` — `TextureSource` was accepted by the public API but hard-coded to `std::fs::read`.** The trait is now plumbed through `doc::build_tree` → `texture::emit_textures_and_videos`. Read errors propagate via `Result` instead of being swallowed by `unwrap_or_default()`; FBX now matches the GLB pipeline's error contract for missing/unreadable textures.
- **`texture.rs` — texture cache keyed by `PathBuf` alone.** The same image used in two different slot kinds (e.g. both `DiffuseColor` and `NormalMap`) would have shared one `Texture` whose `MogenSlot` custom prop could only carry one slot name. Cache key is now `(PathBuf, &'static str)`.
- **`nodes.rs` — `ScalingMax` declared `Vector3D` but emitted one `f64`.** Now uses `push_prop_vec3` with `[0,0,0]`.
- **`doc.rs` — `AmbientColor` declared `ColorRGB` but emitted one `f64`.** Same fix.
- **`mesh.rs` — UV-layer gate was `!uvs.is_empty()`.** Switched to `Mesh::has_uvs()` which also requires `uvs.len() == positions.len()`. A length-mismatched UV array could have produced a malformed `LayerElementUV`.
- **`skin.rs` — multi-mesh skins merged vertex indices from separate geometries into a single shared Cluster**, silently corrupting weights for any skin bound to more than one mesh. Now emits one Skin Deformer + per-joint Cluster set per `(skin, bound geometry)` pair so cluster `Indexes` never cross geometry boundaries.

## P2 (4)

- **`doc.rs` — `FrontAxisSign` was 1 but we declare -Z forward; should be -1.** Matches Blender's exporter for the Y-up / -Z-forward configuration.
- **`mesh.rs` — polygon-terminator `!c` on `tri[2] as i32` silently corrupted the polygon stream for indices ≥ 2³¹** (the high bit is reused for the negate-end marker). Now returns an explicit `anyhow::Error` before the cast, surfacing the issue at export time.
- **`material.rs` — `DiffuseFactor` was set to `base_color.a`**, but in FBX `DiffuseFactor` scales `DiffuseColor` *brightness* (a material with alpha=0.5 would have appeared half as bright). `DiffuseFactor` now stays at 1.0 and alpha rides on `TransparencyFactor = 1 - α`.
- *(rolled into the `texture.rs` cache fix above)*

## P3 (4)

- **`mod.rs` — doc comment claimed `write_fbx_with_options` streams.** Replaced with an accurate description (build to memory then `BufWriter` to disk; `fbxcel` needs random-access seeks for end-offset back-patching so true streaming isn't viable).
- **`main.rs` — `--out` arg help text was stale.** Now mentions the `.fbx` default when `--format fbx` is set.
- *(other P3 items were nits absorbed into the touched files)*

## Test coverage strengthened

- Skin test now uses non-identity inverse-bind matrices and asserts `Transform`/`TransformLink` translation components are negations of each other (catches regressions that copy one into both).
- Skin test asserts the Cluster→joint OO direction explicitly *and* that the reverse direction (joint→Cluster) is **not** present.
- Animation test verifies all three keyframe ticks (0, 0.5s, 1.0s), not just the last, and that `KeyAttrFlags` is 0x4 for Linear.
- New `step_interpolation_emits_constant_key_attr_flag` covers the `KEY_ATTR_FLAGS_CONSTANT` path.
- New `rotation_track_emits_euler_degrees` covers the quat→Euler XYZ conversion (90° around Y → ~90 on Y curve, ~0 on X/Z).
- New `empty_scene_produces_valid_fbx` covers the no-nodes / no-materials path.

## Verification after fixes

- `cargo test -p mogen-export --features fbx` → 42/42 (lib 18 + fbx 10 + lights 4 + textures 10).
- End-to-end `mogen build examples/arm.mog -o arm.fbx` → 964.5 KiB FBX with 18 Skin Deformers, 306 Clusters (18 meshes × 17 joints), 5 AnimStacks, 108 AnimCurves, all parsing cleanly via an independent JS-based FBX 7.4 binary walker (separate code path from the `fbxcel`-based reader used in unit tests).

krazyjakee (approved):

oldschoola:
Have not tested this yet, will test it today.

krazyjakee:
Visually confirmed in blender

PR #38, opened 2026-05-10
Add streaming LLM response support with real-time status updates

krazyjakee:
## Summary
This PR adds streaming support for LLM responses, enabling real-time progress feedback during code generation. Users now see fine-grained status updates ("writing geometry", "authoring animation", etc.) as the model emits tokens, rather than waiting for the entire response to complete.

## Key Changes

- **New streaming infrastructure**:
  - Added `stream_sse.rs`: Minimal SSE (Server-Sent Events) reader for both Gemini and OpenAI streaming responses
  - Added `stream_status.rs`: Status tracker that observes cumulative response text and emits human-readable progress labels ("thinking…", "writing header", "choosing materials", etc.) and geometry node counts

- **Provider streaming implementations**:
  - `gemini.rs`: Added `stream_generate()` method that hits `:streamGenerateContent?alt=sse` endpoint (API-key auth only; OAuth falls back to non-streaming)
  - `openai.rs`: Added `stream_generate()` method with `stream: true` and `stream_options.include_usage: true` request parameters

- **Repair loop integration**:
  - `repair.rs`: Added `on_chunk` callback field to `RepairConfig` for per-frame progress updates
  - `provider.rs`: Added `stream_generate()` dispatcher that routes to provider implementations or falls back to non-streaming

- **Studio UI integration**:
  - `llm.rs`: Integrated streaming with 250ms throttling to prevent excessive UI re-layouts while maintaining responsive status transitions
  - Uses `StreamStatus` behind a `Mutex` to track and emit only when progress changes

- **CLI/batch updates**:
  - Updated all CLI commands (`generate.rs`, `modify.rs`, `animate.rs`, `repair.rs`, `bench.rs`) to pass `on_chunk: None` in `RepairConfig` initialization

- **Test coverage**:
  - Added comprehensive streaming tests in `mock_server.rs`:
    - `gemini_stream_accumulates_deltas_and_invokes_on_chunk_each_frame`: Validates delta accumulation and callback invocation
    - `gemini_stream_targets_the_stream_endpoint_with_sse_alt`: Ensures correct endpoint targeting
    - `openai_stream_accumulates_deltas_and_surfaces_tail_usage`: Validates OpenAI's usage tail frame handling
  - Added unit tests for `StreamStatus` and SSE parsing

## Implementation Details

- **Graceful degradation**: Providers without streaming support transparently fall back to non-streaming; the callback simply never fires
- **Usage handling**: Gemini emits usage on the final frame; OpenAI uses `include_usage` tail frame — both properly surfaced in the returned `GenerateResponse`
- **Empty delta filtering**: Skips invoking callbacks for empty-string deltas to avoid spurious UI updates
- **Status progression**: DSL authoring stages are tracked in order (meta → material → scene → animation), with later stages overriding earlier ones so the user sees the most recent progress
- **Node counting**: Geometry node count is tracked via substring matching of attribute openers (`(size=`, `(radius=`, `(height=`) for cheap, parse-free progress indication

https://claude.ai/code/session_01MxoYW1qdsuviSjrujdSW2R

### DIFF
--- a/crates/mogen-llm/src/gemini.rs
+++ b/crates/mogen-llm/src/gemini.rs
@@ -845,6 +845,16 @@ struct RawGenerateEnvelope {
 
 #[derive(Debug, Deserialize)]
 struct RawCandidate {
+    // Gemini streams emit several candidate shapes during one response:
+    // text-bearing frames (`content.parts[0].text`), `finishReason`-only
+    // tail frames, and the occasional safety-ratings-only frame. Only
+    // the first carries a `content` block, so this field has to be
+    // tolerated as absent — without `#[serde(default)]` the SSE
+    // accumulator's `from_str` fails on the first non-text frame, sets
+    // `sse_err`, aborts the stream, and the caller sees a half-finished
+    // response. Found in production: a single `finishReason` frame from
+    // gemini-2.5-flash was cutting `modify` off after one token.
+    #[serde(default)]
     content: RawContent,
 }
 
--- a/crates/mogen-llm/src/gemini.rs
+++ b/crates/mogen-llm/src/gemini.rs
@@ -523,7 +523,11 @@ impl GeminiClient {
         if let Some(e) = sse_err {
             return Err(e);
         }
-        if cumulative.is_empty() {
+        // Mirror the OpenAI streaming path: a whitespace-only response
+        // is just as useless as an empty one (no parser anchors land,
+        // the repair loop has nothing to feed back), so treat them the
+        // same.
+        if cumulative.trim().is_empty() {
             return Err(GeminiError::EmptyResponse);
         }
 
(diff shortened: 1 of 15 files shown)


krazyjakee:
Still not working in testing. The LLM runs for a few seconds and then ends as soon as it begins to process the response — Modify in Studio shows "thinking…" briefly, then the progress card disappears with no result landed. Same symptom on both Gemini and OpenAI.

The defensive parsing + non-streaming fallback in the last commits should have caught this, but evidently the stream is returning `Ok` with content that passes the 30-byte length check but still isn't a usable response. Need to investigate further — likely candidates: (1) the stream completes successfully but `parse_edit_blocks` + the rewrite fallback both produce garbage that overwrites the file silently, or (2) something earlier in `run_repair_loop` short-circuits before the validation/repair iteration gets a chance to recover.

PR #50, opened 2026-05-17
feat(building): randomise entrance wall side per seed

krazyjakee:
## Summary

`place_entrances` previously hardcoded the wall order `[South, North, East, West]`, so the first (and for single-entrance buildings, only) entrance always landed on the **south facade**. This PR shuffles that order per-seed so the front door can land on any of the four sides.

- New `entrance_side_order(seed)` helper in `crates/mogen-dsl/src/lower/building/layout/mod.rs` returns a Fisher–Yates shuffle of `[S, N, E, W]` driven by a sub-seed of `cfg.seed`.
- `place_entrances` (emit/openings.rs) consumes the shuffled order; for `entrances=N` it still distributes round-robin, but now through the shuffled list — so a four-door building still gets exactly one entrance per facade.
- `entrance_anchors` (layout/score.rs) uses the same helper, keeping the public-near-entry / secure-away-from-entry scoring term aligned with where doors actually land. The scorer runs before opening placement, so both sides need to derive the same order without sharing mutable state — they do, via the seed-derived sub-state.

## Why

The original layout pipeline treated south as the canonical front for layout scoring and corridor/hotel pivoting. That made every grid/courtyard/office building visually identical from the south. With the entrance side now randomised, multi-seed runs produce visibly varied "front" faces while remaining deterministic for any given seed.

## Implementation notes

- The shuffle uses `attempt_seed(cfg.seed, 0x0E11_7A11)` as a fixed sub-seed, independent of the per-storey RNG stream that drives entrance jitter and interior doors. That means seeds that previously placed the entrance on south still produce identical jitter/door placement; only the wall side rotates.
- Updated test `single_entrance_stays_on_south_front` → `single_entrance_lands_on_one_facade`: asserts the door is on exactly one axis-aligned facade rather than pinning to south.
- The pre-existing `multiple_entrances_fan_out_across_facades` test continues to pass — with `entrances=4` every facade still gets one door regardless of where the shuffle starts.

## Pre-existing test failure (not introduced here)

`windows_on_each_side_never_overlap` fails on this branch **and on master cc9c269** with the same error. The window allocator's `max_per = floor(L/pitch)` over-allocates by one window per segment; the resulting `(j+1)/(n+1)` placement violates the pitch invariant. Verified with a fresh `git worktree` at master. Flagged via spawn_task for follow-up — fix is `max_per = floor(L/pitch) - 1` (min 1). The randomisation just exposes the bug on more facades than south.

## Test plan

- [x] `cargo test -p mogen-dsl --lib building::` — 87/88 pass (one pre-existing failure)
- [x] `cargo test --workspace` — 311/312 pass (same pre-existing failure)
- [x] `single_entrance_lands_on_one_facade` — new test, passes
- [x] `multiple_entrances_fan_out_across_facades` — still passes
- [ ] Visual smoke: run a few seeds through `examples/buildings/*.mog` in MoGen Studio and confirm the front door rotates across facades

### DIFF
--- a/crates/mogen-dsl/src/lower/building/emit/circulation.rs
+++ b/crates/mogen-dsl/src/lower/building/emit/circulation.rs
@@ -120,6 +120,7 @@ pub(in super::super) fn emit_circulation(
                     i,
                     bottom_storey,
                     top_storey,
+                    &layout.storeys,
                     circ_group,
                     graph,
                 )?;
@@ -310,6 +311,7 @@ fn emit_staircase(
     idx: usize,
     bottom_storey: i32,
     top_storey: i32,
+    storeys: &[StoreyPlate],
     parent: NodeId,
     graph: &mut SceneGraph,
 ) -> Result<()> {
@@ -482,7 +484,10 @@ fn emit_staircase(
         ),
     );
     graph.nodes[enclosure.0 as usize].origin = origin.clone();
-    emit_shaft_enclosure(cell_w, cell_d, total_h, enclosure, graph, &origin);
+    let (adj_n, adj_e, adj_s) = shaft_face_adjacencies(&cell.rect, storeys);
+    emit_shaft_enclosure(
+        cell_w, cell_d, total_h, adj_n, adj_e, adj_s, enclosure, graph, &origin,
+    );
     Ok(())
 }
 
@@ -841,10 +846,14 @@ fn emit_elevator(
     // match its own room layout — a single full-height wall could only
     // hold one X column per door (wall_with_holes merges X-overlapping
     // spans), so the per-storey shifts would smear into one giant hole.
+    let (adj_n, adj_e, adj_s) = shaft_face_adjacencies(&cell.rect, storeys);
     emit_shaft_enclosure(
         cell.rect.width(),
         cell.rect.depth(),
         total_h,
+        adj_n,
+        adj_e,
+        adj_s,
         shaft_group,
         graph,
         &origin,
@@ -938,10 +947,19 @@ fn emit_elevator_west_walls(
 /// staircases let the per-storey cell-shared wall from `rooms.rs` carry
 /// the door cutout, while elevators emit per-storey west pieces via
 /// `emit_elevator_west_walls`.
+///
+/// `skip_n` / `skip_e` / `skip_s` extend that same exemption to the other
+/// faces whenever an adjacent `Room` cell already provides closure: the
+/// room's per-cell wall carries any door cutout the BFS placed there, and
+/// a solid shaft_wall on top would shadow the cutout (visually hiding
+/// the door slab and adding a trimesh collider that blocks passage).
 fn emit_shaft_enclosure(
     cell_w: f32,
     cell_d: f32,
     total_h: f32,
+    skip_n: bool,
+    skip_e: bool,
+    skip_s: bool,
     group: NodeId,
     graph: &mut SceneGraph,
     origin: &Option<std::path::PathBuf>,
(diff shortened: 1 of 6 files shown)


krazyjakee:
## Review findings — `feat/random-entrance-wall-side`

Reviewed: code quality, structure, test coverage, and 3D geometry (entrance positions verified across seeds 1–42; GLB validity checked via `inspect`).

A fix commit has been pushed to the branch: `73e0694`.

---

#### 🔴 HIGH — Window allocator pitch violation (pre-existing, now broadly exposed)

**File:** `crates/mogen-dsl/src/lower/building/emit/openings.rs` → `allocate_windows`

The PR correctly identifies this bug and flags it for follow-up, but since randomising the entrance side means windows now appear on all four facades (not just south), the `windows_on_each_side_never_overlap` test now fails on every run rather than occasionally. This makes it a blocker for merging.

**Root cause:** `max_per = floor(L / pitch)` allows `n` windows spaced at `L/(n+1)`, but the pitch constraint requires `L/(n+1) >= pitch`, i.e. `n <= L/pitch - 1`. Using `floor(L/pitch)` lets one extra window slip through on segments where `floor(L/pitch) > L/pitch - 1`.

**Fix (committed in `73e0694`):**
```rust
// was:
.map(|s| ((s.hi - s.lo) / pitch).floor().max(0.0) as usize)
// fixed:
.map(|s| (((s.hi - s.lo) / pitch).floor() as i64 - 1).max(0) as usize)
```

Result: **88/88 tests pass** (was 87/88).

---

#### 🟡 MEDIUM — `shaft_face_adjacencies` aggregates across all storeys for a full-height shaft wall

**File:** `crates/mogen-dsl/src/lower/building/emit/circulation.rs` → `shaft_face_adjacencies`

The function returns `adj_s = true` if *any* storey has a Room adjacent on the south face. `emit_shaft_enclosure` then skips `shaft_wall_s` for the **entire shaft height**. If a storey at the top of a multi-storey building has no room adjacent to the stair's south face, that level's south face will be open — the room wall only exists at the lower storey height.

In practice this is benign for `office-core` (the corridor spans all storeys consistently), and the two regression tests use a controlled seed. However, for `style="grid"` or `style="bsp"` with `floors_above >= 2` and `staircases >= 1`, different storeys can have different room adjacencies, which could expose a gap.

**Suggestion:** Either document the invariant ("circulation cells always have a consistent Room neighbour across all storeys for corridor-bearing styles") or make the per-face skip per-storey rather than for the full shaft. No fix committed — this would need broader test coverage first and the behavior is correct for the tested configurations.

---

#### 🟡 MEDIUM — Stale comment in `pick_door_tree_root` (fixed in `73e0694`)

**File:** `crates/mogen-dsl/src/lower/building/emit/openings.rs`

The comment read "anchor on the south-midpoint baseline" which implied south is still canonical. Updated to "floorplate's south-midpoint as a stable fallback" so it's clear the choice is an implementation detail, not a semantic claim about entrance direction.

---

#### 🟢 LOW — Minor code duplication: `per_side` computation

The 5-line pattern:
```rust
let count = cfg.entrances.max(1) as usize;
let order = entrance_side_order(cfg.seed);
let mut per_side: [usize; 4] = [0; 4];
for i in 0..count { per_side[i % 4] += 1; }
```
appears identically in both `place_entrances` (`openings.rs`) and `entrance_anchors` (`score.rs`). Not a correctness issue since both paths execute the same logic against the same `cfg`, but a helper (e.g. `entrance_distribution(cfg) -> ([WallSide; 4], [usize; 4])`) would make any future change to the distribution logic apply atomically. Not fixed — the change is small enough that the duplication is tolerable, and extracting the helper would cross the `emit` / `layout` module boundary in a way that requires more thought.

---

### Visual verification

Entrance positions confirmed across 8 seeds (1, 2, 3, 4, 5, 7, 10, 42):

| seed | entrance face |
|------|--------------|
| 1    | south        |
| 2    | east         |
| 3    | south        |
| 4    | north        |
| 5    | south        |
| 7    | south        |
| 10   | north        |
| 42   | north        |

GLB validity (`mogen inspect`) confirmed for seeds 1, 2, 4 (single-storey grid) and seed 2 (office-core, 3 storeys). Shaft walls on the office-core model correctly show `shaft_wall_n` + `shaft_wall_e` present, `shaft_wall_s` absent (corridor sits south of the stair on seed=2).

The golden-test failures (`bezier_fender`, `coil_spring`, etc.) are pre-existing on `master` (stub golden files, 130 bytes) and are not introduced by this PR.

---

### Summary

The core randomisation logic is correct: `entrance_side_order` is a proper Fisher–Yates shuffle driven by a sub-seed independent of all per-storey RNG streams; the scorer and emitter share the same helper so predictions and placements are always consistent. The `door_w` threshold tightening (1.1× → 1.0×) is clearly motivated and safe for the stair-entry case. The shaft-wall adjacency fix is architecturally sound for the supported styles.

The one code change needed before merging was the window allocator cap — now fixed.


---

PR #52, opened 2026-05-18
Bundle LODs (3 stages) + imposter under one export option

krazyjakee:
## Summary

- Adds `ExportOptions::bundle_lods_and_imposter` (Studio checkbox: "Bundle LODs (3 stages) + imposter") that does two things in one pass:
  - Generates 3 simplified LOD meshes per non-skinned source mesh (50% / 25% / 12% triangle targets via `meshopt::simplify`) and attaches them to the source node through the `MSFT_lod` glTF extension, with `extras.MSFT_screencoverage` thresholds [0.5, 0.25, 0.1, 0.03] so Godot 4's importer swaps stages by distance.
  - Bakes a single scene-wide imposter spritesheet from 8 yaw views via the existing `mogen-render` headless GL stack, embeds the PNG into the GLB BIN chunk, and emits an unlit billboard quad tagged `"imposter"` in `extras.tags` so the companion `godot-mog` runtime can pick the correct cell at sample time.
- mogen output stays plain glTF — the angle-picking shader is a `../godot-mog` responsibility, not this PR.
- New cargo features `lod` (pulls `meshopt`) and `imposter` (pulls `mogen-render` + `image`) gate the heavy deps. Both on by default for desktop; the wasm crate keeps them off.
- Minor: `mogen-render`'s `with_gl_context` now opts into `any_thread` on Linux so the bake works from the Studio's build worker thread (and from `cargo test`), and `Renderer::render_to_pixels` now takes `[u8; 4]` so the imposter can clear to transparent. Thumbnail callers are unaffected (`ThumbnailOptions::bg` stays `[u8; 3]`).

## Test plan

- [x] `cargo build --workspace` — clean (default features include `lod` + `imposter`).
- [x] `cargo test -p mogen-export --test lod_imposter` — three integration tests:
  - `option_off_does_not_emit_lod_or_imposter` — baseline writer output unchanged when flag is off.
  - `option_on_emits_msft_lod_on_source_node` — 3 LOD meshes appended, `MSFT_lod` + `MSFT_screencoverage` stamped on source node, `extensionsUsed` advertises `MSFT_lod`. Skips on headless CI (no `DISPLAY`).
  - `option_on_emits_imposter_node_at_scene_root` (#[ignore]'d — needs a real display): asserts the imposter root node exists, is tagged `"imposter"`, and `KHR_materials_unlit` is advertised. Verified locally on X11.
- [x] `cargo test --workspace` — passes outside two pre-existing unrelated failures on master (`mogen-dsl::lower::building::tests::single_storey::windows_on_each_side_never_overlap` and `mogen-llm::provider::tests::from_env_returns_missing_key_for_blank_cloud_provider`).
- [ ] Manual: open Studio, toggle the new checkbox in Build GLB, inspect the resulting `.glb` shows the source meshes + LOD orphans + imposter root.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1338,6 +1338,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "float-cmp"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b09cf3155332e944990140d967ff5eceb70df778b34f77d8075db46e4704e6d8"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "foldhash"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2395,6 +2404,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "meshopt"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e01e77ead21976b3a9f01ec1724f766923da74f0726364e2b0f425658935d71a"
+dependencies = [
+ "bitflags 2.11.1",
+ "cc",
+ "float-cmp",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "metal"
 version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2504,9 +2525,11 @@ dependencies = [
  "fbxcel",
  "glam",
  "image",
+ "meshopt",
  "mogen-core",
  "mogen-dsl",
  "mogen-geom",
+ "mogen-render",
  "oxipng",
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,11 @@ serde_json = "1"
 glam = { version = "0.30", features = ["serde"] }
 image = { version = "0.25", default-features = false, features = ["png", "jpeg"] }
 oxipng = { version = "9", default-features = false, features = ["parallel"] }
+# meshoptimizer Rust bindings. Used by `mogen-export`'s LOD pass to generate
+# simplified index buffers via quadric edge collapse. Wraps a small C++ lib
+# behind `cc`; doesn't cross-compile to `wasm32-unknown-unknown`, so wasm
+# builds keep the `lod` feature disabled.
+meshopt = "0.6"
 # `import` is the simple high-level API for reading .glb/.gltf into typed
 # accessors. Default features pull in the `image` crate for embedded textures
 # (we strip materials/textures on import, but keeping defaults avoids one more
--- a/crates/mogen-export/Cargo.toml
+++ b/crates/mogen-export/Cargo.toml
@@ -4,7 +4,7 @@ version.workspace = true
 edition.workspace = true
 
 [features]
-default = ["textures", "textures-optimize", "merge"]
+default = ["textures", "textures-optimize", "merge", "lod", "imposter"]
 # Texture packing. Splits cleanly across two layers so wasm builds can
 # embed images without pulling in `image`/`oxipng` (which depend on C libs
 # that don't cross-compile to wasm32):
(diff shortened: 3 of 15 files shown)


krazyjakee:
## Review findings & fixes — commit `0dd1556` on `claude/inspiring-wozniak-nFpeD`

All fixes were applied and pushed. Tests: 97 passed, 0 failed (headless CI, default features).

---

### 🔴 High — Test gap: LOD writer output was never verified in headless CI

**File:** `crates/mogen-export/tests/lod_imposter.rs`

`option_on_emits_msft_lod_on_source_node` had a dead code path: when the `imposter` feature is compiled in (the default) and there's no display, the test returned early with a comment saying "exercise just the LOD half" — but the code never did that. On every standard headless CI run the MSFT_lod attachment was completely untested.

**Fix:** Added `ExportOptions::bundle_lods` (LOD generation without the imposter bake, no display required) and a new `lod_only_emits_msft_lod_on_source_node` integration test that always runs. Retained the original combined test for the display-available path; it now falls through to the runtime error branch instead of returning early before even attempting the export.

---

### 🟡 Medium — LOD generation cloned full vertex buffers that the writer immediately discarded

**File:** `crates/mogen-export/src/lod.rs`

`build_lod_meshes` returned `Vec<Mesh>` with `.positions.clone()`, `.normals.clone()`, `.uvs.clone()` on every LOD stage. The writer never read those fields — it reused the original source node's accessor indices (`pos_acc`, `nrm_acc`, `uv_acc_opt`) and only called `push_indices` on the LOD mesh. For a mesh with N vertices this allocated 3× N vertices of heap data per LOD stage, all immediately garbage.

**Fix:** Changed the return type to `Vec<LodIndices>` (`type LodIndices = Vec<u32>`). Added `push_indices_raw(bin, views, accessors, indices, vertex_count)` to `accessor.rs` so the writer can emit the LOD index accessor using the source mesh's vertex count for the u16/u32 threshold decision.

---

### 🟠 Low — `emit_imposter` docstring described unimplemented sampler reuse

**File:** `crates/mogen-export/src/imposter.rs`

The function doc said *"Re-uses the table's sampler slot 0 when present so we don't double-emit the standard linear/repeat sampler"* but the code always pushes a fresh `CLAMP_TO_EDGE` sampler entry. The behaviour is correct (the imposter atlas needs `CLAMP_TO_EDGE` to prevent cell-bleeding; the shared PBR sampler uses `REPEAT`), but the comment described non-existent optimisation logic.

**Fix:** Replaced the stale sentence with an accurate description of why a dedicated sampler is always created.

---

### 🟠 Low — Hardcoded `8` in writer node extras instead of constant

**File:** `crates/mogen-export/src/writer.rs`

`"view_count": 8` in the imposter root node's extras was a magic number. `VIEW_COUNT` already existed in `imposter.rs` but was module-private.

**Fix:** Promoted `VIEW_COUNT` to `pub(crate)` in `imposter.rs` and replaced the literal in the writer.

---

### 🔵 Trivial — Callsite comment said "XZ extent", function computes XY

**File:** `crates/mogen-export/src/imposter.rs`

The inline comment above the `scene_xy_extent` call said *"Sized to the scene's XZ extent"* (the ground footprint in a Y-up coordinate system), but the function computes XY (width × height) — the correct dimensions for sizing a vertical front-facing billboard.

**Fix:** Updated comment to "XY footprint" to match the function name and implementation.

---

### ✅ Items reviewed and found correct

- **`meshopt::simplify_decoder` API call** — `[f32; 3]` implements `DecodePosition` in meshopt 0.6.x; calling `simplify_decoder` with `&source.positions` is the idiomatic pattern and produces correct simplification.
- **UV flip on billboard quad** — V=0 at top of quad / top of texture is consistent with glTF's top-left UV origin.
- **`any_thread(true)` on the winit event loop** — safe because the headless loop never pumps OS events; it exits as soon as the render closure returns.
- **`CLAMP_TO_EDGE` imposter sampler** — correctly prevents yaw-0/yaw-N−1 bleed at the seam; right choice over the shared `REPEAT` PBR sampler.
- **MSFT_lod orphan node strategy** — LOD nodes correctly sit outside `scene.roots`; importers that don't recognise the extension see the unmodified original scene. Thresholds slice correctly via `SCREEN_COVERAGE[..=lod_node_ids.len()]`.
- **`scene_xy_extent` sizing logic** — computing `max(X_range, Y_range)` for a square billboard's half-extent is correct: it guarantees the quad contains the widest visible silhouette from any yaw.


---

PR #55, opened 2026-05-19
studio: add read-only Connections navigator to the Selected panel

krazyjakee:
Surfaces the symbolic reference graph (parent/child hierarchy, attach
in/out, conform, material, joints, clip tracks, skin) for the selected
node, with every node-to-node link clickable to retarget the selection.
Read-only by design — clicks only move the selection, no source is
mutated, so it works for array/CSG/imported nodes too.

https://claude.ai/code/session_017LD4CFZ6GgbED7443474V4

### DIFF
--- a/crates/mogen-studio/src/app/ui_panels/selected/add_attr.rs
+++ b/crates/mogen-studio/src/app/ui_panels/selected/add_attr.rs
@@ -1,0 +1,88 @@
+//! Schema-driven "Add attribute" picker.
+//!
+//! Several useful common attributes (`anchor`, `from`/`to`, the relative-
+//! placement family, `lod`, `gap`, `role`, `tags`) have no dedicated widget,
+//! so they were previously undiscoverable from the inspector — you had to
+//! know the DSL and edit text. This lists the ones that are valid for the
+//! selected kind (per the validator's own schema tables) and not already
+//! present, and splices a sensible default via the normal span-aware
+//! `SetAttrCanonical` path so the write is undoable and round-trips cleanly.
+//!
+//! Scope is deliberately the placement/metadata attrs without other UI:
+//! transforms (grid), `mat` (picker), `collider`/`cast_shadow` (checkboxes),
+//! deform knobs (Deform section) and the kind geometry params (grid) all
+//! have first-class editors already and are excluded to avoid two controls
+//! fighting over one attribute.
+
+use eframe::egui;
+
+use crate::edit::get_attr;
+use crate::viewer::PendingEdit;
+
+/// Attrs we offer, with a default literal chosen to be schema-valid so the
+/// inserted line builds (the user then edits the value). Relative-placement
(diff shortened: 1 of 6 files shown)


krazyjakee:
## Review findings — commit `4189d53`

All issues below were addressed in the follow-up commit pushed to this branch. 21 new unit tests added; all 236 tests pass.

---

### 🔴 High — Silent data loss on wrong-arity `size` vector

**File:** `geom_params.rs`

The original `vec_row` fallback for vectors whose component count didn't match the expected arity was:
```rust
VecField::Nums(_) => vec[default; arity],
```
This replaced the *entire authored value* with primitive defaults. A user who had `size=[1, 2]` (mismatched arity for a `box`, e.g. from a partial edit) would see `[1.0, 1.0, 1.0]` in the inspector and the first drag on any axis would overwrite `[1, 2]` with `[1.0, 1.0, 1.0]` — silent data loss with no warning.

**Fix:** Use the authored values, padded/truncated to the expected arity with the default:
```rust
VecField::Nums(mut n) => { n.resize(arity, default); n }
```

---

### 🟡 Medium — `mod` declarations not in alphabetical order

**File:** `mod.rs`

`node_problems` was declared after `light_editor`, breaking the alphabetical ordering used everywhere else in this directory.

**Fix:** Re-ordered to alphabetical (`light_editor` before `node_problems`).

---

### 🟡 Medium — `from`/`to` in `CANDIDATES` conflicts silently with `pos_shadows`

**File:** `add_attr.rs`

`from` and `to` in the Add-attribute picker are intended as *path coordinate* attrs (`from=[0,0,0] to=[0,1,0]` for sweeps/extrudes). However, those same names appear in `transform_grid.rs`'s `pos_shadows` delete list. This means the first transform-grid drag after the user adds one of them will delete it — the drag emits `pos=[…]` with `delete: ["x","y","z","from","to"]`.

The interaction is correct by construction (the shadow-delete mechanism pre-dates this PR), but the conflict was invisible. **Fix:** Added a comment above `CANDIDATES` documenting the interaction so future editors understand why path-coord `from`/`to` can disappear after a gizmo drag.

---

### 🟢 Low — Missing unit tests for new pure-logic helpers

**Files:** `node_problems.rs`, `geom_params.rs`, `transform_grid.rs`

All three new modules contained non-trivial pure functions with no test coverage:

- `intersects(a, b)` — half-open span overlap check (subtle off-by-one at adjacent spans)
- `field` / `vec_field` — DSL expression classifier (absent / literal / locked)
- `expr_attr` — parametric-expression detector for transform channels

**Fix:** Added 21 unit tests covering the above, including:
- Adjacent spans do *not* intersect (end is exclusive)
- `Field::Locked` for `$param` refs and arithmetic expressions
- `VecField::Locked` when any component is non-literal
- `expr_attr` skips a literal attr and catches a later param ref in the same scan


---

PR #56, opened 2026-05-23
feat(cli): add `mogen mcp` subcommand exposing the CLI as an MCP server

krazyjakee:
## Summary

Adds `mogen mcp`, a stdio Model Context Protocol server that exposes every existing CLI subcommand as an MCP tool. Designed to be launched by an MCP client (Claude Desktop, a custom client, etc.); a separate client implementation will follow.

Built on the official `rmcp` 1.7 SDK with `schemars`-derived JSON input schemas. **23 tools** total:

- **Pipeline**: `build`, `parse`, `check`, `dump_scene`, `inspect`, `thumbnail`
- **LLM-driven**: `generate`, `modify`, `animate`, `repair`, `textures`
- **Ops**: `update`, `bench`
- **MoGHub**: `moghub_whoami`, `moghub_discover`, `moghub_info`, `moghub_download`, `moghub_comments`, `moghub_comment`, `moghub_like`, `moghub_unlike`, `moghub_notifications`, `moghub_publish`

## Why this design

Each tool call **spawns the same `mogen` binary as a child process** and captures its stdout + stderr, rather than calling command functions in-process. Three reasons:

1. Several commands call `std::process::exit(1)` on validation errors (`check`, `auth`, `moghub`). In-process invocation would kill the MCP server itself.
2. Many commands `println!` to stdout. In stdio MCP mode our stdout is the JSON-RPC transport — any stray print would corrupt the protocol. The child writes into a pipe we capture.
3. Future CLI changes (new flags, new subcommands) need no MCP-side refactor — the args forward verbatim.

Tracing is forced to stderr (`with_writer(std::io::stderr)`) so nothing leaks onto stdout. Default log filter is `warn` when `RUST_LOG` is unset, for the same reason.

On non-zero child exit, the tool returns a `CallToolResult` with `isError=true` and stderr in the content — so the calling LLM gets the diagnostic, not an opaque RPC failure.

## Test plan

- [x] `cargo build -p mogen` — clean.
- [x] `cargo test -p mogen` — 2 unit + 17 golden tests, all pass.
- [x] Smoke test via hand-rolled JSON-RPC client: `initialize` → `notifications/initialized` → `tools/list` returns all 23 tools with correct schemas → `tools/call` for `check` on a passing `.mog` returns success with the expected `ok: ... (7 nodes, 2 materials)` summary.
- [x] Error paths: `check` on `tests/broken/clip_bad.mog` returns `isError=true` with JSON diagnostics; `build` on a missing path returns `isError=true` with the OS error from stderr.
- [x] `serverInfo` reports `name: "mogen"`, `title: "MoGen"`, `version: 0.1.5` (not `rmcp` — `Implementation::from_build_env()` reads `CARGO_*` env vars at the rmcp crate's compile site, so we build the struct ourselves).
- [ ] Wire up a real MCP client (forthcoming, separate PR).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,6 +646,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "chrono"
+version = "0.4.44"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
+dependencies = [
+ "iana-time-zone",
+ "num-traits",
+ "serde",
+ "windows-link",
+]
+
+[[package]]
 name = "chunked_transfer"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -928,6 +940,40 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f27ae1dd37df86211c42e150270f82743308803d90a6f6e6651cd730d5e1732f"
 
 [[package]]
+name = "darling"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "25ae13da2f202d56bd7f91c25fba009e7717a1e4a1cc98a76d844b65ae912e9d"
+dependencies = [
+ "darling_core",
+ "darling_macro",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9865a50f7c335f53564bb694ef660825eb8610e0a53d3e11bf1b0d3df31e03b0"
+dependencies = [
+ "ident_case",
+ "proc-macro2",
+ "quote",
+ "strsim",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "darling_macro"
+version = "0.23.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac3984ec7bd6cfa798e62b4a642426a5be0e68f9401cfc2a01e3fa9ea2fcdb8d"
+dependencies = [
+ "darling_core",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "debugid"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


krazyjakee:
## Code Review — `mogen mcp` subcommand

Reviewed `crates/mogen/src/commands/mcp.rs` (923 lines), `Cargo.toml`, `cli/cmd.rs`, and `main.rs`. Build clean; all existing unit tests pass. Two fixes pushed to `claude/inspiring-wozniak-17bCD`.

---

### Findings (ranked by severity)

#### 🔴 Medium — API keys forwarded as process arguments (security, not fixed)

`api_key` and `zai_api_key` are appended to the child `mogen` argv via `push_opt(args, "--api-key", p.api_key)`. That makes them visible in `/proc/<pid>/cmdline` and `ps aux` output on the host for the lifetime of the child process. Any local user with `/proc` access can read them.

This mirrors the existing CLI behaviour (users already pass `--api-key` on the terminal), but in an MCP context the keys originate from a client configuration file or secret store — the user may not realise they're briefly exposed in the process table.

**Recommended follow-up**: strip `api_key`/`zai_api_key` from the argv before spawning and inject them as environment variables using `Command::env("GEMINI_API_KEY", ...)` / `Command::env("ZAI_API_KEY", ...)` instead. This matches the env-var paths those commands already support.

---

#### 🟡 Low — `server_info` double-assignment in `get_info()` (fixed)

`info.server_info` was assigned `Implementation::from_build_env()`, then `.name` and `.version` were mutated on that value, then `info.server_info` was assigned again via `.with_title().with_description()`. The logic was correct but read as if the first three lines were dead code.

Fixed by introducing a local `let mut si` so the intent is clear in one read-through. See commit `320cf2e`.

---

#### 🟡 Low — No unit tests for arg-building helpers (fixed)

`push_opt`, `push_flag`, and `push_path_opt` were exercised only indirectly through the golden test suite. A flag name typo (e.g. `"--no-cach"`) would silently pass all existing tests. Added six direct unit tests covering both Some/true and None/false branches for each helper. See commit `320cf2e`.

---

#### ℹ️ Info — `rt-multi-thread` tokio feature is heavier than needed

`run()` uses `Builder::new_current_thread()`, which only requires the `rt` feature. `rt-multi-thread` (the thread-pool scheduler) is never activated. This adds compile time and ~100 KB to the binary.

Not fixed because changing tokio features without verifying rmcp's internal requirements carries regression risk. Worth revisiting once the MCP path has integration tests.

---

### No issues found

- **File size**: 923 lines — under the 1 000-line threshold; no split needed.
- **3D model / visual changes**: none — this PR wraps the CLI in a protocol adapter and does not touch geometry, export, or rendering code.
- **Golden tests**: 5 pre-existing failures (`bezier_fender`, `coil_spring`, `heightfield_terrain`, `metaball_blob`, `wave_water`) — unrelated to this PR.
- **Subprocess design rationale** (`process::exit` + stdout contamination) is sound and well-documented.
- **Error surfacing** (`CallToolResult::error` with stderr content) correctly follows the MCP spec so the calling LLM receives diagnostics rather than an opaque RPC failure.


---

PR #57, opened 2026-05-23
Claude/inspiring wozniak 17b cd

krazyjakee:


krazyjakee:
## Review findings & fixes

Reviewed the MCP server implementation (`crates/mogen/src/commands/mcp.rs` + `Cargo.toml`). All issues found have been fixed in commit `ac9f070` on `claude/inspiring-wozniak-SdJsE`. No 3D model / visual changes — this PR only adds a CLI dispatch layer.

---

### 🔴 HIGH — Positional args not guarded by `--` (potential clap mis-parse)

**Files:** `mcp.rs` — `generate`, `modify`, `animate`, `moghub_comment`

`prompt` and `body` are user-controlled strings passed as positional CLI args to the child process. If they start with `--` (e.g. a prompt like `"--no-cache or maybe not"`), clap in the child process tries to look up the string as a registered flag, fails, and returns a confusing error rather than running the intended command.

**Fix:** Reorder each affected tool so named flags are assembled first, then `"--"` terminates option parsing, then the positional string(s) come last:
```
["generate", "--out", "foo.glb", …, "--", "--weird-prompt-text"]
["modify",   "--out", "foo.glb", …, "--", "file.mog", "the prompt"]
["animate",  "--out", "foo.glb", …, "--", "file.mog", "the prompt"]
["moghub", "comment", "--server", "…", "--", "user/slug", "body text"]
```
Four regression tests added to verify `--` precedes the user string in each case.

---

### 🟡 MEDIUM — `rt-multi-thread` Tokio feature pulled in unnecessarily

**File:** `crates/mogen/Cargo.toml`

```toml
# before
tokio = { version = "1", features = ["rt-multi-thread", "macros", "io-std", "process", "sync"] }
# after
tokio = { version = "1", features = ["rt", "macros", "io-std", "process", "sync"] }
```

The MCP server unconditionally uses `Builder::new_current_thread()`, so the multi-thread executor is never instantiated. `rt-multi-thread` compiles in the full thread-pool scheduler unnecessarily. Replaced with `rt`.

---

### 🟡 MEDIUM — Duplicate stale comment in `get_info`

**File:** `mcp.rs` lines 898–903

Two nearly-identical comments explained the same `Implementation::from_build_env()` override. The older, slightly inaccurate version was left in alongside the corrected one. Removed the first copy.

---

### 🔵 LOW — Missing doc comments on `BenchArgs` fields, `MoghubRefOnly.reference`, `MoghubCommentArgs.reference`

Every other struct in the file has doc comments on all fields — these were missed. Doc comments propagate into the generated JSON Schema, which MCP clients use to describe tools to the connected LLM. Added matching comments.

---

### 🔵 LOW — Test coverage gaps

Added 4 new tests:
- `llm_common_push_maps_all_set_fields` — verifies all `LlmCommon` fields emit the correct flag name
- `llm_common_push_omits_none_fields` — verifies empty `LlmCommon` emits nothing
- `generate_args_puts_separator_before_prompt` — regression for the `--` fix
- `moghub_comment_args_puts_separator_before_body` — regression for the `--` fix

All 10 tests pass (`cargo test -p mogen commands::mcp`).

---

krazyjakee:
Closing as duplicate — the MCP server feature shipped via #56 (commit ee688d9) and the review fixes from this branch's comment were applied separately. This branch is now stale and conflicting with master.

Issue #59, opened 2026-05-23
Studio: AI-driven scene generation wizard


krazyjakee:
## Summary

Add a multi-step "Scene Wizard" in MoGen Studio that turns a single user prompt into a fully populated, visually checked scene. The wizard chains existing pipeline pieces (`mogen-llm` text generation, `mogen-llm/textures` image generation, the DSL build pipeline, and the Studio viewer) with new orchestration on top, so the user can go from "a cosy reading nook" to a finished `.glb` scene with correctly placed, individually-generated props.

The defining preset is **isometric, densely populated** — the goal is convincing, lived-in scenes, not minimal compositions.

## Pipeline

1. **Prompt → scene brief.** User enters a free-form prompt. The wizard applies a fixed preset (isometric camera, dense object population, lighting/background hints) and asks the LLM for a high-level scene brief.
2. **Scene → object manifest.** The LLM analyses the brief and produces a comprehensive list of objects in the scene. Each entry includes name, role (hero / filler / decor), rough size, and target world-space position + orientation.
3. **Per-object image + prompt.** For each manifest entry, generate (a) a short focused text prompt and (b) a reference image via the existing Gemini image pipeline in `crates/mogen-llm/src/textures/`.
4. **Per-object model generation.** For each entry, run a model-generation pass that takes the image + prompt and produces a `.mog` module (reusing `generate` / `modify` flows in `crates/mogen-llm`). The output must carry a **position guide** — anchor point, up axis, footprint AABB, and any required connectors — so the assembly step can place it deterministically.
5. **Scene assembly.** Compose the per-object modules into a single parent `.mog` (likely via `use "…"` of each generated module) with positions/rotations from the manifest + position guides. Build to `.glb`.
6. **Visual check — per object.** For each generated model, render a thumbnail in the Studio viewer and feed it back to the LLM with the original per-object prompt; flag/regenerate any that fail the visual check.
7. **Visual check — full scene.** Screenshot the assembled scene from the isometric preset, hand it to the LLM with the original user prompt, and let it propose positional corrections. Apply corrections by editing the assembly `.mog` via `crates/mogen-studio/src/edit.rs` (span-preserving). Loop until stable or a max-iteration cap is hit.

## UX

- A wizard window with one panel per stage and an explicit "Next" gate, so the user can inspect/edit the object manifest, regenerate a single object, or skip the auto-correction step.
[...]

PR #63, opened 2026-05-25
studio: AI Scene Wizard for multi-stage scene generation (#59)

krazyjakee:
Adds File → Scene Wizard, a multi-step pipeline that turns a single
prompt into a fully populated isometric `.glb` scene:

  Prompt → Brief → Manifest → References → Per-object .mog
         → Assemble + Build → Per-object review → Scene review

Each stage is gated behind an explicit user click so the manifest,
generated images, and per-object outputs can be inspected and edited
between LLM calls. Per-stage outputs (state.json, references/, objects/,
scene.mog, scene.glb, scene_thumb.png) are persisted under
`<project>/wizard/` so a partial run survives Studio restart and reruns
skip already-completed work.

The full-scene visual-check loop renders an isometric screenshot
(30° pitch / 45° yaw, shared with the viewer via `iso_camera()`),
asks the LLM for position corrections as strict JSON, and applies the
deltas through the span-preserving `edit::set_attr` helper so unrelated
formatting in the assembly file stays intact. Capped at 3 correction
iterations per run.

Reuses existing infrastructure end-to-end — `LlmClient` /
`generate_with_repair` for every text call, `ImageClient` for reference
images, `pipeline::compile + write_glb_with_source_and_options` for the
build, `viewer::CaptureRequest` (new `WizardThumb` kind) for the
isometric screenshot. No new HTTP path, no new caching layer.

16 unit tests covering manifest parsing, fence stripping, correction
application, and assembly composition.


krazyjakee:
## Review findings and fixes

Reviewed `ui_wizard.rs`, `wizard/pipeline.rs`, `wizard/corrections.rs`, `wizard/assemble.rs`, and `wizard/state.rs`. All fixes are committed to `claude/inspiring-wozniak-mSzwi` (commit `b3083fb`).

---

### 🔴 HIGH — Bug: ReviewScene "← Back" jumped to wrong stage

**File:** `ui_wizard.rs` → `draw_review_scene_stage`

The Back button dispatched `WizardAction::BackToObjects` which calls `wizard_back_to(Stage::Objects)`, skipping the intervening Assemble and ReviewObjects stages entirely. The user was dropped back two stages instead of one.

**Fix:** Added `BackToReviewObjects` variant + dispatch (`wizard_back_to(Stage::ReviewObjects)`); the Back button now dispatches `BackToReviewObjects`.

---

### 🔴 HIGH — Bug: `group_targets_object` name-prefix block was a no-op

**File:** `wizard/corrections.rs`

The function contained an `if` block that checked whether a group name starts with the sanitised object id prefix — but the block body was empty (just a comment). The function always fell through to the `use`-based child scan regardless, making the prefix check dead and the comment misleading. Any group whose `use` child matched the object id would be rewritten, with no actual prefix guard.

**Fix:** Removed the empty block; the function now cleanly expresses the actual intent — match only on the inner `use` target:
```rust
fn group_targets_object(group: &Node, object_id: &str) -> bool {
    group.children.iter().any(|child| {
        child.kind == "use" && child.name.as_deref() == Some(object_id)
    })
}
```

---

### 🟡 MEDIUM — Dead code suppressed with `#[allow(dead_code)]`

Three items were carried but never read:

| Location | Item | Note |
|---|---|---|
| `wizard/assemble.rs` | `relative_objects_dir` | Never called anywhere |
| `wizard/pipeline.rs` | `ast_has_module` | Never called anywhere |
| `wizard/pipeline.rs` | `WizardRunConfig::{claude_code_path, zai_base_url}` | Stored in config struct but never read by any worker; the client constructor is called before thread spawn in `ui_wizard.rs` with values read directly from `self.settings` |

**Fix:** Removed all three, along with the now-unused `Path` import in `assemble.rs` and the `mogen_dsl::ast::Node` import in `pipeline.rs`.

---

### 🟡 MEDIUM — `ui_wizard.rs` exceeded 1 000-line threshold (was 1 575 lines)

**Fix:** Converted to a directory module. The 11 `draw_*_stage` functions + `apply_wizard_message` (≈680 lines) were extracted to `ui_wizard/stages.rs`. Both resulting files are under 1 000 lines:
- `ui_wizard.rs` — 897 lines (session types, all `impl MogenStudioApp` wizard methods, background-worker spawners)
- `ui_wizard/stages.rs` — 716 lines (stage draw functions + message fold)

---

### 🟡 MEDIUM — Stage strip showed "→" after the final "Done" step

**File:** `ui_wizard.rs` → `draw_stage_strip`

The separator was appended unconditionally after every stage label, including the last one.

**Fix:** Only emit "→" when `i < STAGES.len() - 1`.

---

### 🟢 LOW — Missing test: coincidental group-name prefix should not trigger rewrite

**File:** `wizard/corrections.rs`

No test covered the case where two objects share a sanitised-name prefix (e.g. `chair` and `chair_extra`). Added `group_with_matching_prefix_but_different_use_is_not_rewritten` which verifies that only the group whose inner `use` matches the corrected id is rewritten; the coincidentally-prefixed sibling is left untouched.

All 17 wizard tests pass after the changes.

---

PR #64, opened 2026-05-26
refactor: DRY duplicated helpers across DSL, export, and Studio

krazyjakee:
## Summary

Five small refactors that eliminate duplicated logic the codebase audit flagged. No behavior changes.

- **highlight.rs** — Replace the manually-maintained `KEYWORDS` array (which carried a "keep in sync with `KNOWN_KINDS`" comment and was missing 15+ recent primitives) with a direct reference to `mogen_validate::KNOWN_KINDS`. A small `ENUM_VALUE_KEYWORDS` list keeps the bare enum tokens (`directional`, `tile`, `hinge`, …) that aren't node kinds. New primitives now auto-highlight.
- **DSL attribute helpers** — Delete three identical `str_attr` / `string_attr` / `string_or_ident` locals across `attach.rs`, `skin_lower.rs`, and `anim_lower.rs` and route all callers through the existing `Node::attr_string` method.
- **mogen-export `accessor.rs`** — Add a `push_view_and_accessor` helper and named constants for the glTF magic numbers (`TARGET_ARRAY_BUFFER`, `COMPONENT_F32`, etc.). Each of the 10 `push_*` functions now ends in a single struct-literal call instead of ~10 lines of `BufferView` + `Accessor` boilerplate.
- **Studio `transform_grid.rs`** — Extract a `drag_triple` helper for the X/Y/Z `DragValue` pattern, deduplicating translate/rotate/scale rows.

Net: 212 insertions / 220 deletions across 6 files.

Two audit findings were dropped after verification:
- The "centralize `PrimitiveKind` enum in `mogen-core`" proposal turned out to require enum-ifying the AST node kind across many crates — way out of scope for the actual lockstep-edit hazard, which was the single stale comment in `highlight.rs` (fixed above).
- The "unify Gemini OAuth retry scaffolding" proposal was based on bad analysis — `oauth_post_with_retry` and `oauth_post_image_with_retry` already share `oauth_post_inner` with a header-applier `fn` pointer.

## Test plan

- [x] `cargo check --workspace` — clean
- [x] `cargo test --workspace` — all tests pass except 5 pre-existing failures on master (`provider::tests::from_env_returns_missing_key_for_blank_cloud_provider` racing on env vars; 4 `edit::tidy::tests` formatting tests producing an extra blank line). Verified those same tests fail on the unmodified `master` HEAD before any of these changes.
- [ ] Manual smoke: open Studio, confirm syntax highlighting still paints all primitive kinds and the transform inspector still emits canonical `pos=`/`rot=`/`scale=` writes.

### DIFF
--- a/crates/mogen-dsl/src/anim_lower.rs
+++ b/crates/mogen-dsl/src/anim_lower.rs
@@ -48,7 +48,7 @@ fn compose_with_rest_pose(clip: &mut Clip, graph: &SceneGraph) {
     }
 }
 
-use crate::ast::{Node, Value};
+use crate::ast::Node;
 
 /// Lower a top-level or scene-level `joint` node.
 pub fn lower_joint(node: &Node, graph: &mut SceneGraph) -> Result<()> {
@@ -59,7 +59,8 @@ pub fn lower_joint(node: &Node, graph: &mut SceneGraph) -> Result<()> {
     let kind = joint_kind_attr(node)?;
     let axis = node.attr_vec3("axis").unwrap_or(Vec3::Y);
     let limits = node.attr_pair("limits");
-    let pivot_ref = string_or_ident(node.attr("pivot"))
+    let pivot_ref = node
+        .attr_string("pivot")
         .ok_or_else(|| anyhow!("joint \"{name}\" requires pivot=\"<node_name>\""))?;
     let pivot = find_node_scoped(graph, &pivot_ref, node.use_id)
         .ok_or_else(|| anyhow!("joint \"{name}\" pivot \"{pivot_ref}\" is not a scene node"))?;
@@ -133,9 +134,10 @@ fn lower_track(
     };
 
     // Direct node track: caller must supply an explicit property.
-    let prop = string_or_ident(node.attr("prop"))
+    let prop = node
+        .attr_string("prop")
         .ok_or_else(|| anyhow!("direct-node track requires prop=\"translation|rotation|scale\""))?;
-    let property = match prop.as_str() {
+    let property = match prop {
         "translation" | "pos" => TrackProperty::Translation,
         "rotation" | "rot" => TrackProperty::Rotation,
         "scale" => TrackProperty::Scale,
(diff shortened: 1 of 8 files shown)


krazyjakee:
## Review: Findings & Fixes

Reviewed the 6 changed files. All existing tests pass (the 5 golden failures + 4 `edit::tidy` failures are pre-existing on `master`, confirmed by running the same tests on the base). No behavior regressions.

Three findings pushed as a follow-up commit (`7b71657`), ranked by severity:

---

### 🔴 Medium — Wrong return-type description in `drag_triple` doc comment
**File:** `crates/mogen-studio/src/app/ui_panels/selected/transform_grid.rs:35`

The doc comment said _"Returns `true` if any of the three changed"_ but the actual return type is `Option<u8>`. Callers of the scale row depend on the `Some(i)` axis index to apply the proportional-scale logic — an incorrect doc sends readers to the wrong mental model.

**Fix:** Updated to _"Returns `Some(i)` where `i` is the axis index (0=X, 1=Y, 2=Z) that last changed, or `None` if no axis changed."_

---

### 🟡 Low — Missing regression test for the main PR benefit (new keyword auto-highlight)
**File:** `crates/mogen-studio/src/highlight.rs`

The stated goal of the `highlight.rs` change is that new primitives added to `KNOWN_KINDS` (after the old hand-written list went stale) are now highlighted automatically. But no test checked whether any of those 15+ new kinds actually painted as keywords. The existing `highlights_keyword_over_identifier` test used `"box"`, which was in both the old list _and_ `KNOWN_KINDS`, so it wouldn't have caught a regression.

**Fix:** Added `highlights_kinds_absent_from_old_keyword_list` — tests `chamfered_box`, `metaball`, `coil`, `extrude`, and `heightfield`, all of which were in `KNOWN_KINDS` but absent from the old `KEYWORDS` array.

---

### 🟢 Low — Style inconsistency in `String` construction from `attr_string`
**Files:** `crates/mogen-dsl/src/attach.rs:192–196`, `crates/mogen-dsl/src/skin_lower.rs:297–298`

`parent`/`child` in `attach.rs` used `.map(String::from)` while `socket`/`plug` in the same function used `.unwrap_or("…").to_string()`. `skin_lower.rs` also used `.map(String::from)`. Callers in `anim_lower.rs` (not touched here) use neither because `attr_string` returns `&str` directly and those sites don't need ownership.

**Fix:** Normalised to `.map(str::to_string)` everywhere an owned `String` is needed from `attr_string`, matching the idiom used by the rest of the module.

---

No other issues found. The `push_view_and_accessor` refactor is correct — `views.len() - 1` and `accessors.len() - 1` are safe because we push before reading `len`. The `KNOWN_KINDS.contains(&word)` linear scan is equivalent in cost to the old `KEYWORDS.contains(&word)` and both arrays are small enough that this is not a hot-path concern.


---

PR #65, opened 2026-05-26
feat(blob): true SDF + surface-nets organic-mass primitive

krazyjakee:
## Summary

Adds a `blob { ... }` container that meshes implicit-field children via fast-surface-nets — the path the existing `metaball` / `union(smooth=k)` nodes were too approximate to take. Concave junctions (smooth eye sockets, nostrils, smooth-subtracted cavities) and large blend radii now work correctly where the vertex-fillet union approximation broke down.

This unlocks LLM-driven generation of organic anatomical forms — skulls, hearts, organs, faces, monster heads — that previously came out as blocky robot-droid assemblies of `sphere + box + cone` parts.

### Core primitive (`mogen-geom`)
- **`sdf`** — closed-form SDFs for sphere / ellipsoid / box / rounded_box / capsule / cylinder / torus, polynomial smin/smax, and a field evaluator that mixes Add and Subtract ops in order.
- **`isosurface`** — wrapper around `fast-surface-nets` v0.2 with padded AABB, defensive voxel cap, planar XZ-projected UVs.
- **`subdivide`** — Loop subdivision post-pass with proper boundary handling, smoothed UVs, recomputed normals.
- 15 new unit tests across the three modules.

### DSL (`mogen-dsl`)
- New `blob` kind in `lower/blob.rs`. Children must be implicit-field primitives; each accepts the usual `pos` / `rot` / `scale` plus an optional `op="subtract"` (synonyms `sub` / `carve`) to carve a smooth cavity rather than add mass. At least one additive child is required.
- `subdivide=N` as a general post-pass on any mesh-producing kind (primitives, CSG ops, blob). Capped at 3 (≤ 64× tris). Lives in `helpers::apply_subdivide`, called from the three mesh-set sites.

### LOD
- `blob.resolution` goes through `scaled_count` (linear, like primitive `segments`).
- `subdivide` goes through `scaled_subdivisions` (log2 offset, like `icosphere.subdivisions`).
- A global `lod_scale=0.5` drops a 120k-tri skull to ~30k; 0.25 to ~8k.

### Validator (`mogen-validate`)
- `blob` added to `KNOWN_KINDS`, `subdivide` and `op` added to `GEOMETRY_COMMON_ATTRS`, attribute types declared.

### Examples (`examples/nature/`)
- `skull.mog` — canonical anatomy example using the carve-from-a-block approach (one head-shaped ellipsoid + brow + chin protrusions, then subtractive mandible undercut / cheek recesses / temple flattening / eye sockets / nasal / mouth slit).
- `heart.mog` — minimal blob (3 additive + 1 subtract).

### Docs (`docs/dsl.md`)
- New "Implicit-field blob" and "Mesh subdivision" sections.
- `blob` row in the primitives table.
- `lod_scale` line updated to mention `blob.resolution` and any `subdivide=`.

### LLM system prompt (`mogen-llm`)
- KINDS_REFERENCE row spelling out blob's attrs, allowed children, `op=subtract`, and the *"stack of bumps reads as robot; blob of ellipsoids reads as bone"* framing.
- PREAMBLE rule 8 nudge for single-mass anatomical forms, with two explicit anti-patterns surfaced by real LLM regen failures:
  - Do NOT split a skull into separate cranium + jaw blobs and `attach` them (each blob is its own SDF field — surfaces don't blend across blob boundaries, the result is two disconnected pieces).
  - For complex anatomy prefer the **carve-from-a-block** pattern over **stack-of-bumps** (the latter produces a balloon head with a tiny face stuck on).
- `"a human skull"` fewshot at realistic 0.15 m head size, verbatim from the example file.

Prompt-cap test comments updated with the rationale trail: `cacheable_block` 31_000 → 32_000 (+blob KINDS_REFERENCE row, ~700 B); `inline_block` 26_500 → 30_000 (+PREAMBLE nudges + skull fewshot, ~3.5 KB across three iterations as production runs surfaced the anti-patterns).

## Why

User can now generate things like *"a human skull"* and get an actual skull (cranium dome, deep eye sockets, nasal cavity, mouth slit, defined jawline and chin from multiple angles) rather than a sphere balanced on a box with a cone for a nose. The carve-from-a-block insight applies beyond skulls — same pattern works for monster heads, faces, hearts, brains, organs.

## Test plan

- [x] All new modules unit-tested (15 tests in `mogen-geom::{sdf, isosurface, subdivide}`).
- [x] All 36 `mogen-llm::prompt::tests` pass (prompt assembly, cap tests, byte-stable output).
- [x] Full workspace `cargo test` passes except the pre-existing `mogen-llm::provider::tests::from_env_returns_missing_key_for_blank_cloud_provider` env-var test that fails on `master` too.
- [x] `examples/nature/skull.mog` and `heart.mog` validate clean.
- [x] Skull builds to 120k tris at LOD 1.0, ~30k at 0.5, ~8k at 0.25 (verified with `lod_scale (value=X)` wrapper).
- [x] Visually verified: skull reads as a skull from front, 3/4, side, above, and front-high views (eye sockets, brow, cheekbones, mandible, chin all visible).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1336,6 +1336,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7360491ce676a36bf9bb3c56c1aa791658183a54d2744120f27285738d90465a"
 
 [[package]]
+name = "fast-surface-nets"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2aef9e8d9e33f51d8120a2e997a819f18239ad7c2735e353ade79208a9a1c442"
+dependencies = [
+ "glam 0.29.3",
+ "ndshape",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1635,6 +1645,12 @@ dependencies = [
  "log",
  "xml-rs",
 ]
+
+[[package]]
+name = "glam"
+version = "0.29.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8babf46d4c1c9d92deac9f7be466f76dfc4482b6452fc5024b5e8daf6ffeb3ee"
 
 [[package]]
 name = "glam"
@@ -2655,7 +2671,7 @@ dependencies = [
 name = "mogen-anim"
 version = "0.1.7"
 dependencies = [
- "glam",
+ "glam 0.30.10",
  "mogen-core",
 ]
 
@@ -2663,7 +2679,7 @@ dependencies = [
 name = "mogen-core"
 version = "0.1.7"
 dependencies = [
- "glam",
+ "glam 0.30.10",
  "serde",
 ]
 
@@ -2672,7 +2688,7 @@ name = "mogen-dsl"
 version = "0.1.7"
 dependencies = [
  "anyhow",
- "glam",
+ "glam 0.30.10",
  "mogen-anim",
  "mogen-core",
  "mogen-geom",
@@ -2687,7 +2703,7 @@ version = "0.1.7"
 dependencies = [
  "anyhow",
  "fbxcel",
- "glam",
+ "glam 0.30.10",
  "image",
  "meshopt",
  "mogen-core",
@@ -2705,7 +2721,8 @@ version = "0.1.7"
 dependencies = [
  "anyhow",
  "earcutr",
- "glam",
+ "fast-surface-nets",
+ "glam 0.30.10",
  "gltf",
  "manifold-csg",
  "mogen-core",
@@ -2717,7 +2734,7 @@ version = "0.1.7"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
- "glam",
+ "glam 0.30.10",
  "image",
  "mogen-core",
  "mogen-dsl",
@@ -2766,7 +2783,7 @@ name = "mogen-render"
 version = "0.1.7"
 dependencies = [
  "anyhow",
- "glam",
+ "glam 0.30.10",
  "glow 0.14.2",
  "glutin",
  "glutin-winit",
@@ -2789,7 +2806,7 @@ dependencies = [
  "eframe",
  "egui",
  "egui_glow",
- "glam",
+ "glam 0.30.10",
  "glow 0.14.2",
  "image",
  "mogen-core",
@@ -2920,6 +2937,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee6cda3051665f1fb8d9e08fc35c96d5a244fb1be711a03b71118828afc9a873"
 dependencies = [
  "jni-sys 0.3.1",
+]
+
+[[package]]
+name = "ndshape"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "975bce586ad637e27f6dc26ee907c07050686a588695bfd64b7873a9d48a700c"
+dependencies = [
+ "static_assertions",
 ]
 
 [[package]]
--- a/crates/mogen-dsl/src/lower.rs
+++ b/crates/mogen-dsl/src/lower.rs
@@ -1,4 +1,5 @@
 mod anim;
+mod blob;
 mod branch;
 mod building;
 pub(crate) mod connector;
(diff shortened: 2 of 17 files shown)


krazyjakee:
## Review findings & fixes — commit `1adb398`

Reviewed code quality, file size/structure, test coverage, and visual output for the `blob` + `subdivide` feature. Four issues found and fixed; ranked by severity.

---

### 🔴 HIGH — `evaluate_field` silently drops subtract-before-add children

**File:** `crates/mogen-geom/src/sdf.rs` · `evaluate_field`

The evaluator iterated children in declaration order, starting with `field = +∞`. A `Subtract` child processed before any `Add` child called `smax(+∞, -d, k) = +∞` — an identity that leaves the field unchanged. The add that follows would set the field correctly, but without the subtraction applied. In practice the skull and heart examples always list additive children first, but the DSL has no enforcement of that ordering, so any user who writes:

```
blob {
  sphere (op=subtract, ...)   // carver first
  ellipsoid (...)              // main mass
}
```

would get a blob with the carver silently ignored.

**Fix:** two-pass evaluation — accumulate all `Add` children with `smin` first, then apply all `Subtract` children with `smax`. Child ordering is now irrelevant. Regression test `subtract_before_add_still_carves` added.

---

### 🟡 MEDIUM — `world_aabb` recomputes `self.inv.inverse()` on every call

**File:** `crates/mogen-geom/src/sdf.rs` · `BlobChild::world_aabb`

The previous struct stored only `inv: Mat4`. `world_aabb()` recovered the forward transform via `self.inv.inverse()` at runtime. `blob_to_mesh` calls this twice per build (two-pass AABB), so a skull with 12 children ran 24 matrix inversions that could be free. Matrix inversion also accumulates floating-point error with each round-trip.

**Fix:** `BlobChild` now stores both `xform: Mat4` (forward) and `inv: Mat4` (inverse), both set in `new()`. `world_aabb` uses `self.xform` directly.

---

### 🟡 MEDIUM — `decompose_scale` is a trivial passthrough wrapper

**File:** `crates/mogen-geom/src/sdf.rs`

```rust
fn decompose_scale(m: &Mat4) -> (Vec3, glam::Quat, Vec3) {
    let (s, r, t) = m.to_scale_rotation_translation();
    (s, r, t)
}
```

The function returns the exact tuple returned by the stdlib method, used at one callsite that immediately discarded `r` and `t`. The indirection adds noise with no abstraction benefit.

**Fix:** removed; `BlobChild::new` calls `xform.to_scale_rotation_translation()` directly.

---

### 🟢 LOW — Multi-paragraph doc comments on `apply_subdivide` / `SUBDIVIDE_CAP`

**File:** `crates/mogen-dsl/src/lower/helpers.rs`

Both the constant and the function had multi-paragraph docstrings explaining the LOD math in detail. The project convention (CLAUDE.md) is "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."

**Fix:** trimmed to one-line summaries:
```rust
/// Max Loop subdivision rounds for any mesh-producing node (cap is 64× tri count).
pub(super) const SUBDIVIDE_CAP: u32 = 3;

/// Apply `subdivide=N` Loop subdivision to `mesh` (clamped via LOD scale; no-op if absent or 0).
pub(super) fn apply_subdivide(node: &Node, mesh: Mesh) -> Result<Mesh> {
```

---

### Visual review ✅

Built and rendered `examples/nature/skull.mog` (119.6k tris, 747ms) and `examples/nature/heart.mog` (14k tris, 39ms) via the headless thumbnail pipeline.

**Skull (front view):** eye sockets, nasal cavity, mouth slit, brow ridge, and chin protrusion all clearly visible. The carve-from-a-block approach works — the cranium reads as a skull face, not a balloon head. No mesh artifacts, clean normals, watertight output. ✅

**Heart:** two smooth lobes with a visible carved cleft at the top centre, tapered apex below. Clean cardioid silhouette, no artifacts. ✅

One cosmetic note: the default thumbnail yaw (π/4) shows the back of the skull. The front face (yaw=π) is the diagnostic view for this asset, so `skull.mog` might benefit from a `# thumbnail: yaw=3.14` hint comment in the file — but this is not a blocker.

---

### Not changed / out of scope

- Missing validation snapshot tests for blob-specific errors (`tests/broken/blob_no_children.mog` etc.) — left for a follow-up rather than blocking this PR since the validator schema entries are correct and the lowering already returns descriptive bail! errors.
- The 5 pre-existing golden test failures (`bezier_fender`, `coil_spring`, `heightfield_terrain`, `wave_water`, `metaball_blob`) are unchanged by this PR and were failing on `master` before it.


---

PR #66, opened 2026-05-28
Fix stale GlitchTip DSN in mogen-studio crash reporting

krazyjakee:
The crash-reporting DSN in `crates/mogen-studio/src/crash.rs` pointed at a project/key that no longer exists on the redeployed GlitchTip instance — and project id 3 there now belongs to **moghub**, so any events would have been misrouted or rejected.

Point it at the new `mogen-studio` project (id 4, org `cattrallgames`). The rest of the consent-gated init (env opt-out, `crash_reports_enabled` flag, panic capture) is unchanged — this is a one-line DSN fix.

Verified end-to-end against the live GlitchTip project (test event ingested, then deleted).

### DIFF
--- a/crates/mogen-studio/src/crash.rs
+++ b/crates/mogen-studio/src/crash.rs
@@ -10,7 +10,7 @@
 //! [`Settings::crash_reports_enabled`] consent flag — Sentry only initialises
 //! when the user has explicitly opted in.
 
-const DSN: &str = "https://fec046177fb54225805b1c011f783d5d@crash.daccord.gg/3";
+const DSN: &str = "https://9ee8ddd1a1b8436fa31e56cafa33a1bb@crash.daccord.gg/4";
 
 /// True when the environment hard-disables telemetry. Independent of any
 /// saved consent — env-level opt-out wins, so the first-launch prompt should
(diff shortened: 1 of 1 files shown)


krazyjakee:
## Code Review — findings & fixes (ranked by severity)

### 🟡 Medium — Missing test coverage for `telemetry_blocked_by_env` (fixed)

**File:** `crates/mogen-studio/src/crash.rs`

The env-based opt-out path (`MOGEN_DISABLE_TELEMETRY` / `DO_NOT_TRACK`) had zero test coverage. This is a privacy promise to users: if either env-var check ever silently regressed, crash data could be sent when the user explicitly opted out via environment. A regression here would be invisible without tests.

**Fix applied in this review commit (`b5f2e41`):** Added three serialised unit tests covering:
- `MOGEN_DISABLE_TELEMETRY=1` → blocked
- `DO_NOT_TRACK=1` → blocked
- Neither set → not blocked

All three pass (`cargo test -p mogen-studio crash`).

---

### ✅ No issues — DSN change itself

The one-line DSN update is correct:
- Same host (`crash.daccord.gg`), project ID bumped from `3` → `4` (matching the description's `mogen-studio` project)
- `send_default_pii: false` remains in place
- Consent gating (`consented != Some(true)` + env opt-out) is unchanged
- Sentry public DSNs are intentionally embeddable in client code — they are write-only ingestion keys

### ✅ No issues — File size / structure

`crash.rs` is 50 lines (86 with the new tests). No split warranted.

### ✅ No issues — 3D / visual changes

No geometry, rendering, or export code touched — visual review not applicable.


---

PR #70, opened 2026-06-01
ci(release): build Linux binaries on ubuntu-22.04 for older glibc

### LINKED ISSUE #69: Release: Linux binaries require GLIBC 2.38+, fail on Debian 12 / bookworm
## Summary

Prebuilt Linux binaries (both `mogen` and `mogen-studio`) require GLIBC 2.38+ and won't run on older-but-current distros. A user on Debian 12 (GLIBC 2.36, image `node:22-bookworm-slim`) reported:

> These binaries require GLIBC 2.38+. My container current has GLIBC 2.36 (Debian 12)... The mcp route won't work either since the binary won't run inside my container.

This blocks the CLI **and** the `mogen mcp` route inside common container bases. Debian 12 (bookworm) is the current Debian stable, so this affects a large share of containerized users.

## Root cause

The release workflow builds Linux artifacts on `ubuntu-latest` (currently Ubuntu 24.04, GLIBC 2.39) targeting `x86_64-unknown-linux-gnu`, dynamically linked against the runner's glibc:

- `.github/workflows/release.yml:23` — `os: ubuntu-latest`
- `.github/workflows/release.yml:24` — `target: x86_64-unknown-linux-gnu`
- `.github/workflows/release.yml:49,52` — `cargo build` for the gnu target

glibc forward-compat means binaries built against 2.39 require >= that version at runtime.

## Options

1. **Pin to an older runner** — build the Linux target on `ubuntu-22.04` (GLIBC 2.35), which covers Debian 12 (2.36) and most current distros. Lowest-effort, keeps dynamic linking. (Note: GitHub is deprecating ubuntu-20.04 runners, so 22.04 is the practical floor.)
2. **Add a static musl target** — add `x86_64-unknown-linux-musl` artifacts so binaries run regardless of host glibc. Best portability; need to confirm the native Manifold CSG cmake build (`csg` feature) and egui/Studio deps link cleanly under musl.
3. **Build inside an old-glibc container** (e.g. manylinux / Debian 11) during release.

Recommendation: option 1 as a quick fix; option 2 (musl) as the durable solution, at minimum for the `mogen` CLI used by the MCP route.

## Workaround for the reporter

Upgrade the container base, or build from source inside the container (`cargo build -p mogen --release`).

krazyjakee:
## Summary

Pin the Linux release runner to `ubuntu-22.04` (GLIBC 2.35) instead of `ubuntu-latest` (Ubuntu 24.04, GLIBC 2.39).

Prebuilt `mogen` / `mogen-studio` Linux binaries are dynamically linked against the runner's glibc. Built on 2.39, they require GLIBC 2.38+ at runtime and fail on Debian 12 / bookworm (GLIBC 2.36) — blocking both the CLI and the `mogen mcp` route in common container bases like `node:22-bookworm-slim`. GLIBC 2.35 covers Debian 12 and most current distros.

This is Option 1 (the quick fix) from #69. A static musl target remains the durable follow-up.

## Changes

- `.github/workflows/release.yml`: Linux build matrix entry `ubuntu-latest` → `ubuntu-22.04`.

## Notes for reviewers

- The `publish` job stays on `ubuntu-latest` — it only downloads artifacts and runs `sha256sum`, so no glibc-linked binary is produced there.
- `ubuntu-20.04` runners are being deprecated by GitHub, so 22.04 is the practical floor.

Fixes #69

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-latest
+          - os: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
           - os: macos-15-intel
             target: x86_64-apple-darwin


krazyjakee:
## Review findings — ranked by severity

Reviewed the full `release.yml` workflow in context of this change. No bugs found; findings are informational only.

---

### 🟢 LGTM — change is correct

`ubuntu-22.04` ships GLIBC 2.35. Binaries dynamically linked against it require ≥ 2.35 at runtime. Debian 12 (bookworm) ships GLIBC 2.36, so the requirement is satisfied. The `publish` job correctly stays on `ubuntu-latest` — it only downloads artifacts and runs `sha256sum`, no glibc-linked binary is produced there.

The `.deb` asset metadata in `crates/mogen-studio/Cargo.toml` bundles both the `mogen` CLI and `mogen-studio` GUI, which is consistent with the tarball packaging — no gap there.

---

### ℹ️ Informational (Low) — RHEL 9 / CentOS Stream 9 gap remains

RHEL 9, Rocky Linux 9, and AlmaLinux 9 ship GLIBC 2.34 (< 2.35), so they will still reject these binaries. This is not a regression introduced here — it was already broken before the PR. The static musl build mentioned in the PR description as the "durable follow-up" would close this gap entirely.

### ℹ️ Informational (Very Low) — pre-existing: mutable action tag refs

All GitHub Actions in the workflow use mutable version tags (`actions/checkout@v4`, `Swatinem/rust-cache@v2`, `awalsh128/cache-apt-pkgs-action@v1`, etc.) rather than pinned commit SHAs. This is a supply chain security consideration independent of this PR; pinning to SHAs would remove the risk of a compromised tag silently executing arbitrary code in the release pipeline.

---

No fixes required for this PR. The one-line change is minimal, correct, and well-scoped.

---

Issue #68, opened 2026-06-01
Studio: support OpenAI-compatible local LLMs (llama.cpp, LM Studio)


krazyjakee:
## Summary

MoGen Studio's local LLM support is limited to Ollama. Users running other popular local runtimes can't connect:

> local llm support on MoGen Studio is limited to Ollama (no image capability). Personally, I use llama.cpp, but LM Studio is a popular choice too.

The user tried to point the Ollama setting at an OpenAI-style endpoint (`http://localhost:1234/v1`, LM Studio's default), which can't work because the Ollama client uses the native `/api/chat` endpoint rather than `/v1/chat/completions`.

## Context

- An OpenAI-compatible client already exists: `crates/mogen-llm/src/openai.rs` builds `POST {base_url}/chat/completions` (`openai.rs:80`) and supports custom base URLs via `with_base_url()` (`openai.rs:60`). llama.cpp and LM Studio both expose this surface.
- The Ollama client instead hits `POST {base_url}/api/chat` (`crates/mogen-llm/src/ollama.rs:67`), so an OpenAI `/v1` base URL won't route correctly.

## Proposal

Surface a generic "OpenAI-compatible (local)" provider option in Studio settings that lets users set an arbitrary base URL (e.g. `http://localhost:1234/v1` for LM Studio, `http://localhost:8080/v1` for llama.cpp's server) and routes through the existing `OpenAiClient`. This covers llama.cpp, LM Studio, and any `/v1/chat/completions` server in one provider.

## Image capability caveat

The user also noted "no image capability" for local providers. Texture/image generation is wired only to Gemini and Z.ai (`crates/mogen-llm/src/image_client.rs` enum has just `Gemini`/`Zai`; `crates/mogen-studio/src/app/util/textures_run.rs:161`). Local text runtimes generally don't offer image generation, so local mode should degrade gracefully (text DSL generation works; texture generation stays on a cloud provider or is disabled) rather than erroring.

## Related

- #67 (custom Ollama base URL ignored) — adjacent bug in the same config path.

Issue #67, opened 2026-06-01
Studio: custom Ollama base URL setting is ignored


krazyjakee:
## Summary

A user set a custom Ollama base URL (`http://localhost:1234/v1`) in MoGen Studio settings, but generation still hit the hardcoded default and failed:

```
You appear to be offline — Could not reach the provider.
(error sending request for url (http://localhost:11434/api/chat): client error (Connect): tcp connect error: Connection refused (os error 111))
```

The request went to `http://localhost:11434/api/chat` regardless of the configured base URL — the setting does not take effect.

## Root cause

The `ollama_base_url` setting is stored but never threaded through to the client:

- `Settings::ollama_base_url` is defined at `crates/mogen-studio/src/settings.rs:213` (default empty → library default).
- `OllamaClient::with_base_url()` exists (`crates/mogen-llm/src/ollama.rs:53`) and would honor a custom URL, but Studio never calls it.
- `build_run_config()` (`crates/mogen-studio/src/app/llm/spawn.rs:27`) collects `zai_base_url` but has **no** `ollama_base_url` equivalent.
- `LlmRunConfig` (`crates/mogen-studio/src/app/util/llm.rs:20`) has no `ollama_base_url` field.
- `build_provider_client()` (`crates/mogen-studio/src/app/util/llm.rs:158`) falls through to the default `LlmClient::new(Provider::Ollama, api_key)` at line 177, which calls `OllamaClient::new()` → always `DEFAULT_BASE_URL = http://localhost:11434` (`crates/mogen-llm/src/ollama.rs:25,50`).

For contrast, Z.ai (`llm.rs:174`, `spawn.rs:44`) and Claude Code (`llm.rs:171`) both correctly pass their custom base URL/path.

## Fix

Mirror the Z.ai wiring for Ollama: add `ollama_base_url` to `LlmRunConfig`, populate it in `build_run_config()`, and add an Ollama match arm in `build_provider_client()` that calls `OllamaClient::with_base_url()` when the setting is non-empty.

## Note

The user's URL was an OpenAI-style `/v1` base (llama.cpp / LM Studio). Even once wired, the Ollama client hits the native `/api/chat` endpoint, not `/v1/chat/completions` — see the separate local-LLM/OpenAI-compatible feature request.

PR #71, opened 2026-06-01
studio: thread Ollama base URL + add OpenAI-compatible local provider

krazyjakee:
## Summary

Fixes two MoGen Studio provider-wiring issues (#67, #68).

- **Issue #67 — Ollama base URL never threaded.** The `ollama_base_url` setting was persisted but never reached the client, so generation always hit the hardcoded `http://localhost:11434/api/chat`. Per-provider base URLs / binary paths are now bundled into a `ProviderEndpoints` struct that `build_provider_client` consumes; the Ollama arm calls `LlmClient::with_base_url` whenever the setting is non-empty (mirroring the existing Z.ai wiring).

- **Issue #68 — generic OpenAI-compatible (local) provider.** Adds a first-class `Provider::OpenAiCompat` in `mogen-llm` for arbitrary OpenAI-compatible local servers (LM Studio at `:1234/v1`, llama.cpp at `:8080/v1`, etc.). Routes through the existing `OpenAIClient` (`POST {base_url}/chat/completions`). `OpenAIClient` now skips bearer auth when the key is blank so keyless local servers work. Studio gains a matching `ProviderSlot::OpenAiCompat`, settings fields (base URL, key, thinking/fast model), model-field accessors, and a Preferences section with a base-URL input that nudges the user when blank. Text DSL generation works locally; image/texture generation degrades gracefully (no `supports_images` arm → stays on a cloud provider) rather than erroring.

## Why

Both settings existed in the UI/struct but weren't honoured end-to-end. Bundling base URLs into `ProviderEndpoints` avoids pushing extra positional args through the ~6 `build_provider_client` call sites (run_llm, ask, enhance, meta-generate, wizard) and gives one source of truth via `MogenStudioApp::provider_endpoints()`.

## Notes for reviewers

- `OpenAiCompat` is modelled as a real `Provider` (not a Studio-only slot collapsing to `Provider::OpenAI`) because model ids are keyed by `Provider`; collapsing would clobber the user's real OpenAI model id with a local one.
- CLI is unaffected — `From<ProviderArg>` is exhaustive over `ProviderArg`, not `Provider`, so the new variant doesn't force a CLI change.
- Two pre-existing test failures are unrelated to this change and reproduce on the base: `provider::tests::from_env_returns_missing_key_for_blank_cloud_provider` (reads the real `~/.mogen/settings.json`, which has an OpenAI key, so `from_env` resolves it instead of erroring) and 4 `edit::tidy::tests` source-formatting tests (`edit/tidy.rs` is untouched).

## Test plan

- [x] `cargo build --workspace` (clean)
- [x] `cargo test --workspace` — all provider/LLM-wiring tests pass; only the two unrelated pre-existing failures noted above remain
- [ ] Manual: point Studio at a local Ollama with a custom base URL and confirm generation hits it
- [ ] Manual: select "OpenAI-compatible (local)", set base URL to an LM Studio / llama.cpp server, confirm text generation works and textures fall back to the cloud provider


krazyjakee:
## Review findings — PR #71 (Ollama base URL threading + OpenAI-compatible local provider)

Build and full test suite: ✅ clean (`cargo build --workspace`, `cargo test --workspace` — 301 unit + integration tests pass).

No 3D model / geometry changes in this PR — visual review skipped.

---

#### 🔴 High — `provider_keys_round_trip` omitted `OpenAiCompat`

`provider::tests::provider_keys_round_trip` hardcodes the provider list but was missing the new variant. A future rename of `Provider::OpenAiCompat`'s key string would silently pass CI. **Fixed.**

#### 🔴 High — `provider_parse_accepts_aliases` had zero coverage for `OpenAiCompat`

Every other provider in `Provider::parse` had at least one alias exercised in the test. `OpenAiCompat`'s 10+ aliases (lmstudio, llama.cpp, llamacpp, lm-studio, openai-compatible, local-openai, …) were entirely untested. **Fixed** — six aliases added.

#### 🟡 Medium — No `from_env` keyless test for `OpenAiCompat`

`Ollama` has `from_env_succeeds_for_ollama_without_key`; `ClaudeCode` has `claude_code_is_keyless_in_from_env`; `OpenAiCompat` had nothing. A regression in `is_keyless()` would go undetected. **Fixed** — `from_env_succeeds_for_openai_compat_without_key` added.

#### 🔵 Low — `settings.rs` (1069 lines) and `prefs.rs` (1199 lines) exceed the 1000-line threshold

Both files were already over the limit before this PR; the additions are small and in the natural locations (Settings struct fields, Preferences UI section). Splitting either file is a standalone refactor — flagging here for tracking but not blocking this PR.

---

### Fix commit

`0467029` — `test(provider): cover OpenAiCompat in all three existing provider test suites`

Pushed to `claude/dazzling-northcutt-b2f007`.

---

### Code quality notes (no action required)

- **`settings_store::read_api_key`** — the `Provider::ClaudeCode | Provider::OpenAiCompat => return None` arm is technically dead code (the `is_keyless()` guard at the top already returns early), but is required for match exhaustiveness. Same pre-existing situation exists for the `Ollama` arm. No change needed.
- **`openai.rs` keyless-auth change** is clean and correct — trimming before the empty check avoids sending `Authorization: Bearer ` with whitespace-only tokens to public OpenAI.
- **`ProviderEndpoints` struct** cleanly consolidates the two positional `&str` args that were previously threaded through ~6 call sites. Signature stability as new providers are added is now much cheaper.
- **`OpenAiCompat` blank-URL fallback** (falls through to public OpenAI host) is correctly warned in the Preferences UI nudge — acceptable UX tradeoff as documented in the PR description.


---

PR #72, opened 2026-06-01
studio: generate a wizard scene from a single source image

krazyjakee:
## What

Adds an **image-driven mode** to the AI Scene Wizard. Upload one photo and the wizard:

1. Derives the **brief** and **object manifest** from the image (vision input), rather than from a typed prompt.
2. Cuts an **isolated reference for each object** out of the photo via **image-to-image** ("extract just the chair from this scene, isolated on a neutral background").
3. Hands those references to the existing per-object model generator with the usual manifest/brief context.

The typed prompt becomes **optional guidance** when an image is present; with no image, the wizard behaves exactly as before.

## Why

The wizard previously only generated scenes from text. Driving it from a single image lets users reproduce a real scene faithfully — each generated object is grounded in what's actually in the photo instead of a generic text description.

## How

**mogen-llm — input-image conditioning (the image-gen path was text-only):**
- `image.rs`: `build_image_request` now attaches `inline_data` parts **ahead of** the text instruction (Gemini's image-edit ordering); new `GeminiClient::generate_image_with_inputs` with the existing `generate_image` delegating via an empty slice. Both API-key and OAuth flavours carry the parts (`wrap_image_body` preserves `contents` verbatim).
- `zai.rs`: signature-symmetric `generate_image_with_inputs` — `glm-image` is text-to-image only, so inputs are ignored.
- `image_client.rs`: `generate_image_with_inputs` (with spend recording) + `supports_image_input()` for capability gating.

**mogen-studio:**
- `pipeline.rs`: `run_brief` / `run_manifest` / `run_reference_image` accept an optional source image. References use image-to-image when the backend supports it and **fall back to text-only** otherwise.
- `state.rs`: persisted `source_image` field so an image-driven run resumes after a restart.
- `ui_wizard.rs`: Prompt-stage image picker (copies the file into the project dir), image-primary/prompt-optional UI, vision-provider gating with a clear error, and a `read_source_image` helper reusing the existing `guess_image_mime`.

## Reviewer notes
- **Image-to-image needs Gemini** (API key or Antigravity OAuth). With a Z.ai-only image credential, references degrade gracefully to text-only; the brief/manifest still require a vision-capable *text* provider (gated with an explicit error via `Provider::supports_images()`).
- No changes to the assemble/review stages — they consume the per-object `.mog` + thumbnails unchanged.

## Testing
- `cargo test --workspace` — green (added two `build_image_request` tests asserting input-image parts precede the text part).
- `cargo build -p mogen-studio` — clean.
- Text-only wizard path is unchanged (empty `input_images` slice).

### DIFF
--- a/crates/mogen-llm/src/image.rs
+++ b/crates/mogen-llm/src/image.rs
@@ -8,6 +8,7 @@ use serde::Deserialize;
 
 use crate::gemini::{GeminiAuth, GeminiClient, GeminiError};
 use crate::google_oauth;
+use crate::types::ImageInput;
 
 /// Default image model for API-key flows. 2.5 Flash Image ("Nano Banana") is
 /// the cheapest tier on the public `generativelanguage.googleapis.com` surface
@@ -68,6 +69,22 @@ impl GeminiClient {
         prompt: &str,
         seed: Option<u64>,
     ) -> Result<GeneratedImage, GeminiError> {
+        self.generate_image_with_inputs(model, prompt, seed, &[])
+    }
+
+    /// Same as [`Self::generate_image`] but conditions generation on one or
+    /// more input images (image-to-image / editing). The images are attached
+    /// as `inline_data` parts ahead of the text prompt — the shape Gemini's
+    /// image surface uses for "edit this picture" requests. Used by the Scene
+    /// Wizard to cut an isolated per-object reference out of a source photo.
+    /// Pass an empty slice for plain text-to-image.
+    pub fn generate_image_with_inputs(
+        &self,
+        model: &str,
+        prompt: &str,
+        seed: Option<u64>,
+        input_images: &[ImageInput],
+    ) -> Result<GeneratedImage, GeminiError> {
         // Honor the `ImageClient::generate_image` contract that `""` means
         // "use the provider default". Empty model would otherwise produce
         // `/models/:generateContent` and a confusing 404.
@@ -81,7 +98,7 @@ impl GeminiClient {
             GeminiAuth::ApiKey(key) => {
                 // Public API speaks `responseModalities: ["IMAGE"]` and returns
                 // a single JSON envelope.
-                let inner = build_image_request(prompt, seed, Some(&["IMAGE"]));
+                let inner = build_image_request(prompt, seed, Some(&["IMAGE"]), input_images);
                 let url = format!(
                     "{}/models/{}:generateContent?key={}",
                     self.base_url(),
@@ -144,7 +161,7 @@ impl GeminiClient {
                 let project = self
                     .oauth_project_id()
                     .ok_or_else(|| GeminiError::OAuth("missing project id in token bundle".into()))?;
-                let inner = build_image_request(prompt, seed, None);
+                let inner = build_image_request(prompt, seed, None, input_images);
 
                 let candidates: Vec<(String, bool)> =
                     if model == DEFAULT_OAUTH_IMAGE_MODEL || model.is_empty() {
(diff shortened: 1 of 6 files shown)


krazyjakee:
## Code Review — PR #72: wizard scene from image

Reviewed `mogen-llm` (image.rs, image_client.rs, zai.rs) and `mogen-studio` (pipeline.rs, state.rs, ui_wizard.rs). Three issues found and fixed in `claude/inspiring-wozniak-9NiLe` (commit `5ea21a4`). Fixes ranked by severity below.

---

### 🔴 HIGH — `{provider:?}` leaks internal enum names into user-visible error

**File:** `crates/mogen-studio/src/app/ui_wizard.rs`, `start_wizard_brief`

```rust
// Before — shows "Zai", "OpenAiCompat", etc.
session.error = Some(format!(
    "{provider:?} can't read images — switch to a vision provider …"
));

// After — shows "Z.ai (GLM)", "OpenAI-compatible (local)", etc.
session.error = Some(format!(
    "{} can't read images — switch to a vision provider …",
    provider.label()
));
```

`Provider` derives `Debug` but the variants are Rust identifiers (`Zai`, `OpenAiCompat`, `ClaudeCode`). `provider.label()` already exists for exactly this purpose and returns the human-friendly display string used everywhere else in the UI.

---

### 🟡 MEDIUM — Source-image read failure silently becomes a text-only run

**File:** `crates/mogen-studio/src/app/ui_wizard.rs`, `start_wizard_brief`

`has_image` was computed from `source_image.is_some()` *before* calling `read_source_image`. If the file was moved or deleted between picking and clicking "Generate", `has_image` would be `true` (triggering the status line "Reading the source image…") while the worker would receive `None` bytes and fall back to text-only — no error surfaced.

Fixed by reading the file first, deriving `has_image` from the returned bytes, and surfacing a clear error when the path is set but the read fails:

```
"Source image could not be read — it may have been moved or deleted.
 Clear the source image or choose a new one."
```

---

### 🟡 MEDIUM — Source image re-read from disk once per manifest object in a batch

**File:** `crates/mogen-studio/src/app/ui_wizard.rs`, `spawn_reference_worker`

`spawn_reference_worker` called `read_source_image` internally. Since `start_wizard_reference_batch` calls it in a loop (up to 18 objects), the source image file was read from disk up to 18 times per batch run. For a typical 2–5 MB photo this is 36–90 MB of unnecessary I/O.

Fixed by removing the `read_source_image` call from `spawn_reference_worker` and instead:
- `start_wizard_reference_batch`: reads once before the loop, passes bytes+mime to each `spawn_reference_worker` call (cloned per thread — cheaper than disk I/O)
- `start_wizard_one_reference`: reads once before calling `spawn_reference_worker`

`spawn_reference_worker` now documents that the caller is responsible for supplying the image bytes.

---

### ✅ No issues found

- **`image.rs`**: `generate_image_with_inputs` correctly places inline-data parts before the text part; the two new unit tests assert this ordering. The delegation chain (`generate_image` → `generate_image_with_inputs` with `&[]`) is clean.
- **`image_client.rs`**: `supports_image_input` is a simple, correct capability gate; the `generate_image_with_inputs` dispatch is symmetric with the existing patterns.
- **`zai.rs`**: stub-forwarding `generate_image_with_inputs` to `generate_image` and documenting why is the right call — the API has no input-image field.
- **`pipeline.rs`**: The prompt branching for image-driven vs text-only runs in `run_brief`, `run_manifest`, and `run_reference_image` is coherent. The `has_image` guard (`source_bytes.is_some() && source_mime.is_some()`) is consistent throughout.
- **`state.rs`**: `#[serde(default)]` on `source_image: Option<PathBuf>` is correct for backward-compatible deserialization.

### File size note

`ui_wizard.rs` is 2 210 lines (pre-existing; this PR adds ~157 lines). The 1 000-line rule would flag it for splitting, but the refactor is out of scope for this PR. The new helpers (`pick_wizard_source_image`, `clear_wizard_source_image`, `read_source_image`) are well-placed and self-contained.

---

**Tests:** `cargo test --workspace` passes (minus five pre-existing golden regressions unrelated to this change). `cargo build -p mogen-studio` is clean.

---

PR #73, opened 2026-06-01
fix(studio): stop scene wizard runaway image requests

krazyjakee:
## Summary

The Scene Wizard could fire **~2000 Gemini image requests while generating only 3 images**. Root cause: `start_wizard_reference_batch` selected pending objects but never inserted their ids into `running_ref_ids`. The 150ms auto-continue poll (`poll_wizard`) gates on `running_ref_ids.len() < BATCH_SIZE` and re-checks for pending work — since the in-flight set stayed empty, every tick re-spawned workers for the same objects. Image generation takes seconds, so dozens of redundant workers piled up per object before any PNG landed on disk to mark it done.

The object batch path did *not* have this bug because it correctly reserved ids — the two paths had drifted.

## What changed

- Made **selection and reservation a single inseparable operation**: new `claim_pending_batch` helper picks pending objects *and* reserves their ids in the in-flight set in one call, returning the entries to dispatch. Both the reference and object batch paths now use it, so there is no longer a way to claim work without reserving it.
- Extracted shared `reference_missing` / `object_missing` predicates so the poll gate and the claim logic agree on what "pending" means.
- Added a `batch_tests` module (7 tests) that proves the invariant without threads or network.

## Why it's test-proof now

This was an expensive mistake, so the fix is locked behind tests:

- `second_claim_before_completion_is_empty` — the direct regression: claim 3, claim again → empty, no double-issue.
- `poll_gate_sees_no_pending_after_claim` — same guarantee via the real poll probe `next_pending_reference_skipping`, so the 150ms tick can't re-fire.
- capacity caps, in-flight accounting, reclaim-after-completion, and the on-disk `exists()` predicate branches.

Verified the regression test genuinely bites: removing the reservation line makes `second_claim_before_completion_is_empty` fail (`left: 0, right: 3`); restoring it passes. Full suite: **287 passed**.

## Reviewer note

These tests cover the dispatch/poll decision logic. The actual thread-spawning + live Gemini call path is still untested (needs the running app + network), so a single manual wizard run in Studio is still worth doing before trusting it with real quota.

## Test plan

- [x] `cargo test -p mogen-studio` (287 pass)
- [x] Confirmed regression test fails when reservation is reverted
- [ ] Manual: run the Scene Wizard reference stage with "Auto-generate all" and confirm exactly one request per image

### DIFF
--- a/crates/mogen-studio/src/app/ui_wizard.rs
+++ b/crates/mogen-studio/src/app/ui_wizard.rs
@@ -21,6 +21,8 @@ use super::wizard::pipeline::{
 use super::wizard::state::{
     next_pending_review_object, ObjectEntry, ObjectGenResult, Stage, WizardMessage, WizardState,
 };
+use super::wizard::{persist, write_assembly};
+use super::MogenStudioApp;
 
 /// A reference image still needs generating when its PNG isn't on disk yet.
 fn reference_missing(o: &ObjectEntry) -> bool {
@@ -59,16 +61,9 @@ fn next_pending_object_skipping<'a>(
         .find(|o| object_missing(o) && !in_flight.contains(&o.id))
 }
 
-/// Pick up to the remaining-capacity objects whose generated artifact is still
-/// missing and **atomically reserve them** in `in_flight`, returning the
-/// claimed entries to dispatch.
-///
-/// Selection and reservation are deliberately one operation. An earlier bug
-/// selected a reference batch without reserving it, so the 150ms auto-continue
-/// poll saw the in-flight set stay empty and re-spawned workers for the same
-/// objects every tick — thousands of image requests for a handful of images.
-/// Keeping claim+reserve inseparable makes that class of regression impossible
-/// to reintroduce by editing one path and forgetting the other.
+/// Claim pending entries up to `target_concurrency - in_flight.len()` capacity, reserving them
+/// in `in_flight` atomically — selection and reservation must stay one step to prevent the
+/// 150ms poll tick from re-issuing in-flight ids before their artifacts land on disk.
 fn claim_pending_batch(
     manifest: &[ObjectEntry],
     in_flight: &mut HashSet<String>,
@@ -87,8 +82,6 @@ fn claim_pending_batch(
     }
     claimed
 }
-use super::wizard::{persist, write_assembly};
-use super::MogenStudioApp;
 
 /// In-flight pump for the wizard. Holds the message channel and small UI
 /// drafts that don't belong in `WizardState` (per-row name overrides,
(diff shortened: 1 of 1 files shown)


krazyjakee:
## Code Review — findings and fixes (commit 14b8621)

The core fix is correct and the regression test genuinely locks it down. A follow-up commit addressed the issues below. All 288 tests pass.

---

### P1 — Code quality / conventions

**1. Misplaced `use` statements**
`use super::wizard::{persist, write_assembly}` and `use super::MogenStudioApp` were stranded after three function definitions (pre-existing, but the PR inserted another function above them, making it worse). Moved to the top-of-file import block where Rust convention expects them.

**2. Multi-paragraph docstring on `claim_pending_batch`**
CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."* The 10-line `///` block was trimmed to a single line that still captures the invariant (selection + reservation must stay inseparable):
```rust
/// Claim pending entries up to `target_concurrency - in_flight.len()` capacity, reserving them
/// in `in_flight` atomically — selection and reservation must stay one step to prevent the
/// 150ms poll tick from re-issuing in-flight ids before their artifacts land on disk.
```

**3. Multi-line block comments before test functions**
Same convention — 2–3 line `//` blocks before `#[test]` functions trimmed to single lines.

---

### P2 — Test correctness / coverage

**4. Temp-file leak on test panic**
`objects_with_existing_artifact_are_not_claimed` and `object_missing_tracks_mog_path` called `std::fs::remove_file` at the end but not via `Drop`, so a panicking assertion would leave the file behind. Added a `TempFile(PathBuf)` RAII guard with `impl Drop` — cleanup is now guaranteed even on failure.

**5. Missing `in_flight.len()` assertion in `capacity_accounts_for_already_in_flight`**
The test verified the claimed count but not that the pre-existing `o0`/`o1` plus the newly claimed `o2` all end up in `in_flight`. Added `assert_eq!(in_flight.len(), 3)`.

**6. Object batch path not covered through `claim_pending_batch`**
`object_missing_tracks_mog_path` tested the predicate in isolation but never called `claim_pending_batch` with `object_missing`. Added `objects_with_existing_mog_are_not_claimed` to confirm the object batch path filters correctly end-to-end.

---

### Not changed (out of scope for this PR)

- **File size (2 233 lines):** `ui_wizard.rs` is well over the 1 000-line threshold and should be split into focused modules (batch dispatch, poll logic, worker spawning, message handling, UI rendering). That's a standalone refactor — not suitable to bundle with a targeted bug fix.
- **No 3D model / visual changes** in this PR — visual review not required.
- **Manual wizard run** still recommended (noted in the PR description) to confirm exactly one Gemini call per image with "Auto-generate all".

---

PR #75, opened 2026-06-01
fix(studio): stop inspector field keystrokes leaking into code editor

### LINKED ISSUE #74: Studio: typing in right-sidebar fields (position/scale) leaks keystrokes into code editor
## Description

In MoGen Studio, when typing into the numeric fields in the right sidebar (e.g. position / scale inspector fields), the keystrokes are *also* entered into the code editor on the left. Input intended for a sidebar field should be captured by that field only and not propagate to the source editor.

## Steps to reproduce

1. Launch MoGen Studio (`./scripts/run-studio.sh`).
2. Select a node so the inspector appears in the right sidebar.
3. Click into a position or scale field and start typing.

## Expected

Keystrokes go only into the focused sidebar field.

## Actual

The same characters are also inserted into the code editor on the left.

## Notes

Likely a keyboard-focus / input-consumption issue in the egui UI — the sidebar `TextEdit`/`DragValue` isn't capturing focus exclusively, so the code editor also receives the key events. Relevant areas: `crates/mogen-studio/src/app/ui_*` and the code editor widget.

krazyjakee:
## Summary

Fixes #74 — typing into a right-sidebar inspector field (position/scale) also inserted the characters into the code editor on the left.

## Root cause

Each keystroke in an inspector `DragValue` fires `.changed()`, which queues an edit **and** sets `pending_caret` to the node's declaration (`selected/mod.rs`) — a "jump the editor caret to what just changed" feature shared with viewport picks. The editor consumes that `pending_caret` the same frame and unconditionally called `request_focus(editor_id)` (`editor.rs`), yanking keyboard focus out of the field. The next character then landed in the code editor, and the cycle repeated on every keystroke.

## The fix

Only grab editor focus when no *other* widget already owns it:

```rust
let other_focus = ui.ctx().memory(|m| m.focused().is_some_and(|f| f != editor_id));
if !other_focus {
    ui.ctx().memory_mut(|m| m.request_focus(editor_id));
}
```

The focus states distinguish the two paths cleanly:

- **Viewport pick / gizmo release** — the GL viewport uses `Sense::click_and_drag()` (not focusable), so the click surrenders focus to `None` → the guard passes → caret-jump grabs focus exactly as before.
- **Typing in a DragValue** — the field genuinely holds focus → the guard skips `request_focus`, so the field keeps it. The editor caret still moves to the declaration (visual "what changed"); it just no longer steals keystrokes.

## Test plan

- [x] `cargo build -p mogen-studio` compiles clean
- [ ] Manual: select a node, type into a position/scale field, confirm characters stay in the field and don't appear in the code editor, while the editor caret still tracks the edited node
- [ ] Manual: confirm clicking a node in the 3D viewport still jumps the editor caret to its declaration and focuses the editor (unchanged behaviour)

### DIFF
--- a/crates/mogen-studio/src/app/ui_panels/editor.rs
+++ b/crates/mogen-studio/src/app/ui_panels/editor.rs
@@ -317,7 +317,21 @@ impl MogenStudioApp {
                                 CCursor::new(char_idx),
                             )));
                             state.store(ui.ctx(), editor_id);
-                            ui.ctx().memory_mut(|m| m.request_focus(editor_id));
+                            // Only pull keyboard focus into the editor when no
+                            // *other* widget already owns it. A viewport pick /
+                            // gizmo release leaves focus None (the GL viewport
+                            // isn't focusable) so the caret-jump grabs focus as
+                            // before. But an inspector DragValue the user is
+                            // actively typing into holds focus — and that field
+                            // queues this same pending_caret on every keystroke.
+                            // Stealing focus here would route the next character
+                            // into the editor instead of the field (issue #74).
+                            let other_focus = ui
+                                .ctx()
+                                .memory(|m| m.focused().is_some_and(|f| f != editor_id));
+                            if !other_focus {
+                                ui.ctx().memory_mut(|m| m.request_focus(editor_id));
+                            }
                         }
                     }
 
--- a/crates/mogen-studio/src/app/ui_panels/editor.rs
+++ b/crates/mogen-studio/src/app/ui_panels/editor.rs
@@ -317,15 +317,9 @@ impl MogenStudioApp {
                                 CCursor::new(char_idx),
                             )));
                             state.store(ui.ctx(), editor_id);
-                            // Only pull keyboard focus into the editor when no
-                            // *other* widget already owns it. A viewport pick /
-                            // gizmo release leaves focus None (the GL viewport
-                            // isn't focusable) so the caret-jump grabs focus as
-                            // before. But an inspector DragValue the user is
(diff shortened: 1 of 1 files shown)


krazyjakee:
## Review findings

Reviewed commit `35e1b04`. One finding, already fixed in `claude/inspiring-wozniak-hL64K`.

---

### 🟡 Medium — Overly verbose inline comment (fixed)

**File:** `crates/mogen-studio/src/app/ui_panels/editor.rs:320–328`

The PR added an 8-line block comment restating what the PR description already covers in full. Per project conventions comments should only be added when the WHY is non-obvious, and should be kept brief.

Trimmed to 3 lines preserving the two non-obvious invariants:
- the `#74` bug-fix reference
- the viewport-pick / focus-`None` invariant that explains why the guard is safe

```rust
// Don't steal focus from a widget the user is
// actively typing into (fixes #74). Viewport picks
// leave focus None, so the guard still passes.
```

---

### ✅ No other issues found

| Area | Result |
|---|---|
| Logic correctness | ✅ Guard is correct for all three focus states (None / editor / other widget) |
| Find-bar edge case | ✅ Same guard also protects the Ctrl+F field — bonus fix |
| File size | ✅ 439 lines, well under threshold |
| Test coverage | ✅ No unit tests added; egui focus behaviour requires a running egui context and is not feasible to unit-test here. Existing viewer tests unaffected. Manual test plan in the PR description is appropriate. |
| 3D model / visual changes | N/A — focus-only change, no geometry or rendering affected |
| `cargo build -p mogen-studio` | ✅ Clean |

---

PR #77, opened 2026-06-01
fix(geom): stop mesh-merge panicking on non-manifold operands

krazyjakee:
## Summary

Fixes the wasm compile crash reported on [vintage-pc-desk-setup](https://moghub.org/m/krazyjakee/vintage-pc-desk-setup):

```
panicked at crates/mogen-geom/src/csg.rs:128:
CSG operand is not a manifold mesh; check that the primitive is closed: ManifoldStatus(NotManifold)
```

### Root cause

The export-time sibling-merge pass (`merge_sibling_meshes`, forced on in the wasm build) collapses same-material sibling leaves into one CSG union. It gated candidate leaves with `is_closed_manifold`, a cheap check that only verifies every **undirected** edge is shared by exactly two triangles.

The scene's `extrude` panels and `spline_tube` cord are **edge-closed but not manifold** — they have inconsistent/duplicate winding (the `extrude` meshes each had 28 duplicate directed halfedges). They passed the cheap gate, got fed into `union_many` → `build_manifold`, which calls `Manifold::from_mesh_f32(...).expect(...)` and panicked, taking down the whole wasm compile.

### Changes

- **`crates/mogen-geom/src/csg.rs`** — add `is_csg_manifold(mesh)`, the ground-truth predicate that runs the exact import the boolean ops perform, plus `try_union_many` (returns `None` instead of panicking). Also fix a latent debug-assert in `from_manifold_output`/`boolean_inner`: a union mixing a UV operand (5 props) with a non-UV one (3 props) tripped it — UVs are now kept only when **both** operands carry them.
- **`crates/mogen-export/src/merge.rs`** — `is_mergeable` now gates on `is_csg_manifold` of the **baked** mesh (exactly what gets unioned), so inconsistently-wound leaves pass through untouched. `merge_group_meshes` uses `try_union_many` and falls back to the existing passthrough path on failure — belt-and-braces against any fold intermediate that turns non-manifold.
- **`crates/mogen-geom/src/lib.rs`** — export the new functions plus matching no-CSG stubs.

### Verification

- The real `desk2/scene.mog` now merges (460 → 405 nodes) without panicking.
- Added regression tests at both the geom level (`is_csg_manifold_rejects_edge_closed_but_inconsistent_winding`) and the merge level (`inconsistent_winding_sibling_does_not_merge_or_panic`).
- `cargo test --workspace` passes.

### Reviewer note

`is_closed_manifold` is now unused in production (only its own unit tests reference it). Left in place as a documented public geometry predicate; happy to remove it if preferred.

## Test plan

- [x] `cargo test --workspace`
- [ ] Rebuild wasm and confirm vintage-pc-desk-setup compiles on moghub.org


krazyjakee:
## Code Review — findings and fixes

Overall the PR is solid: the root-cause diagnosis is correct, the two-layer defence (gate in `is_mergeable` + fallback in `merge_group_meshes`) is the right shape, and the existing regression tests cover the reported crash well. One contract bug was found and fixed; the rest are informational.

---

### 🔴 [FIXED] Contract violation in `try_union_many` — single non-manifold operand returned `Some` instead of `None`

**File:** `crates/mogen-geom/src/csg.rs`

The function is documented as _"returns `None` the moment any operand fails to import as a manifold"_, but when called with a single-element slice containing a non-manifold mesh the fold body never executes, so `try_boolean` is never called and the bad mesh is returned as `Some(bad_mesh)` unchecked.

```rust
// before
pub fn try_union_many(meshes: &[Mesh]) -> Option<Mesh> {
    let mut it = meshes.iter();
    let first = it.next()?;
    let mut acc = first.clone();          // ← first never validated here
    for m in it { … }
    Some(acc)
}
```

This path cannot be reached through `merge_group_meshes` today (single-element material groups are handled by the `ids.len() < 2` branch before `merge_group_meshes` is called), but it is a public API with a stated contract, and a caller that bypasses the merge plumbing could trigger it.

**Fix pushed (commit `718de46`):** added `try_build_manifold(first)?;` before the fold, plus two new tests:
- `try_union_many_two_valid_manifolds_returns_some` — happy-path coverage that was missing.
- `try_union_many_single_non_manifold_returns_none` — regression for the contract violation.

---

### 🟡 Performance: each merge candidate's mesh is transformed and manifold-imported twice

**Files:** `crates/mogen-export/src/merge.rs` — `is_mergeable` and `merge_group_meshes`

`is_mergeable` calls `transform_mesh` + `is_csg_manifold` (which runs `try_build_manifold` internally) to gate the leaf. Then `merge_group_meshes` calls `transform_mesh` again on the same leaf, and `try_union_many` calls `try_build_manifold` a second time inside `try_boolean`. For a scene like vintage-pc-desk-setup (460 nodes) with many merge candidates this is 2× the transforms and 2× the Manifold imports.

**Not fixed in this PR** — correcting it would require threading the baked mesh out of `is_mergeable` or doing filtering and baking in a single pass through `rebuild_level`, which is a larger refactor. The overhead is bounded and the logic is correct as-is.

---

### 🟢 Informational: `cleanup.unwrap_or(CleanupMode::Standard)` is dead code

**File:** `crates/mogen-export/src/merge.rs` line 196

```rust
let cleanup = merge_here.unwrap_or(CleanupMode::Standard);
for (material, ids) in groups { … }
```

`groups` is only populated inside `if merge_here.is_some()`, so the `.unwrap_or` branch is unreachable in practice. Not a bug — just defensive code that could mislead a reader into thinking `cleanup` might default. No action needed.

---

### 🟢 Informational: `is_closed_manifold` is now production-dead

As noted in the PR description, `is_closed_manifold` is only referenced by its own unit tests now that `is_mergeable` gates on `is_csg_manifold`. Happy to leave it as a documented public geometry predicate (or remove it in a follow-up) — no action needed here.

---

**All tests pass** (`cargo test -p mogen-geom --features csg`, `cargo test -p mogen-export`). The critical fix is in commit `718de46` on this branch.


---

PR #79, opened 2026-06-01
fix(ci): build cargo-deb from source to avoid glibc mismatch

krazyjakee:
## Summary
- The release workflow's `Install cargo-deb` step used `taiki-e/install-action`, which downloads a prebuilt `cargo-deb` v3.7.0 binary linked against `GLIBC_2.39`.
- The Linux build runs on `ubuntu-22.04` (glibc 2.35), so the binary failed with `version 'GLIBC_2.39' not found`, breaking the `.deb` packaging step.
- Switched to `cargo install cargo-deb --locked`, which compiles against the runner's own glibc.

## Why not bump the runner?
Moving to `ubuntu-24.04` would resolve the prebuilt-binary glibc requirement, but it would also raise the glibc floor of the **shipped** `mogen`/`mogen-studio` binaries and the `.deb`, reducing compatibility on older Linux systems. Building on 22.04 keeps that floor low.

## Notes for reviewer
- Only the Linux `cargo-deb` step changed. The Windows `cargo-wix` step still uses `taiki-e/install-action` since the glibc mismatch is Linux-only.
- `cargo install` adds a one-time compile cost (mitigated by the existing `Swatinem/rust-cache`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Install cargo-deb
         if: runner.os == 'Linux'
-        run: cargo install cargo-deb --locked
+        run: cargo install cargo-deb --version 3.7.0 --locked
 
       - name: Package .deb (linux)
         if: runner.os == 'Linux'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,9 +85,7 @@ jobs:
 
       - name: Install cargo-deb
         if: runner.os == 'Linux'
-        uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-deb
+        run: cargo install cargo-deb --locked
 
       - name: Package .deb (linux)
         if: runner.os == 'Linux'


krazyjakee:
## Code Review — findings & fixes (ranked by severity)

### [MEDIUM] Missing version pin on `cargo install` — **Fixed** ✅

**Finding:** `cargo install cargo-deb --locked` installs the latest published version of cargo-deb on every release run. If a new version ships with a packaging bug or a breaking change, all subsequent releases silently produce different (potentially broken) `.deb` files. The original `taiki-e/install-action` step was pinned to v3.7.0.

**Fix applied (commit `28b1598`):**
```diff
- run: cargo install cargo-deb --locked
+ run: cargo install cargo-deb --version 3.7.0 --locked
```

Pins to v3.7.0 — the same version the workflow was using before — but now compiled against the runner's own glibc (2.35) instead of downloaded as a prebuilt binary linked against glibc 2.39.

---

### [LOW] Compile overhead not fully mitigated by `Swatinem/rust-cache` — no code change needed

**Finding:** `Swatinem/rust-cache` caches `~/.cargo/registry` (downloaded source and index) but the compiled `cargo install` binary lives in a temporary build directory that is not restored between runs. Each release run will recompile cargo-deb from scratch (~1–2 min added to the Linux job). This is acceptable for a release workflow but worth being aware of.

**Options if build time becomes a concern:**
- Use `actions/cache` to explicitly cache `~/.cargo/bin/cargo-deb` keyed on the version string.
- Re-evaluate `ubuntu-24.04` once the glibc floor of shipped binaries is reconsidered.

---

### [INFO] Intentional asymmetry between Linux/Windows tool install — no change needed

The PR leaves `cargo-wix` on `taiki-e/install-action` for Windows. This is correct: the glibc mismatch is Linux-specific, and the prebuilt cargo-wix binary works fine on the Windows runner.

---

**No 3D model / geometry changes in this PR — visual review not applicable.**
**No application logic changed — test coverage not applicable.**
**File is 232 lines — no structural split needed.**


---

PR #80, opened 2026-06-01
studio: Tab inserts 2 spaces; debounce inspector transform edits

krazyjakee:
Editor: intercept Tab for all selections (not just multi-line) and insert
two spaces (one indent unit) instead of egui's default literal tab. A
caret or in-line selection inserts a single indent; a multi-line selection
block-indents every covered line. Shift+Tab dedent now matches at 2 spaces.

Performance: inspector position/rotation/scale DragValues fire a fresh
edit every frame of a drag, and drain_viewport_edits compiled immediately
on each one — a full recompile + mesh re-flatten (reloading geometry and
textures from disk) per frame, which hung the app for a second or more
while dragging. Defer the recompile to the existing debounce while the
pointer is mid-drag; keep the immediate compile for discrete commits
(gizmo release, combo picks, typed values, deletes) so the viewport still
updates on the next frame and values can't appear to snap back.

### DIFF
--- a/crates/mogen-studio/src/app/compile.rs
+++ b/crates/mogen-studio/src/app/compile.rs
@@ -275,24 +275,12 @@ impl MogenStudioApp {
                     node_path: self.current_selection_path(i),
                 };
                 self.push_undo(i, undo_before, key);
-                // Inspector DragValues emit a fresh edit on every frame of a
-                // drag, whereas the gizmo and discrete widgets (combo picks,
-                // typed values, deletes) emit once on release. A full
-                // recompile + mesh re-flatten reloads geometry and textures
-                // from disk, so doing it per frame makes the app visibly hang
-                // while a transform field is dragged. While the pointer is
-                // still mid-drag we therefore defer to the debounce —
-                // `drive_compile_debounce` coalesces the burst into a single
-                // recompile once the user pauses (and still fires mid-drag if
-                // they hold without moving for the debounce window).
-                //
-                // On a discrete commit we keep compiling immediately so the
-                // viewport reflects the change on the very next frame.
-                // Gizmo releases are discrete events, not keystroke bursts —
-                // without the immediate compile the preview clears on release
-                // and the new scene only arrives on the next idle tick, which
-                // looks to the user like the action was rejected and the value
-                // snapped back. Deletes are always discrete too.
+                // Inspector DragValues fire every frame of a drag; a full
+                // recompile per frame makes the app visibly hang. Defer to
+                // the debounce while the pointer is held. Gizmo releases and
+                // deletes are discrete — compile immediately so the viewport
+                // updates on the very next frame and values don't appear to
+                // snap back.
                 let mid_drag = ctx.is_using_pointer() && !any_delete;
                 if !mid_drag {
                     self.compile_active();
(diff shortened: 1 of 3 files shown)


krazyjakee:
## Code Review — PR #80 findings & fixes

Pushed one follow-up commit (`70c2586`) with the items below. All 306 tests pass (19 new).

---

### 🔴 Bug (Medium) — block-indent cursor drift with multi-byte INDENT

**File:** `crates/mogen-studio/src/app/indent.rs`

The old loop that adjusted the cursor endpoints after inserting INDENT on each line compared `ls < new_lo` against the *already-mutating* `new_lo`:

```rust
for &ls in &line_starts {
    if ls < new_lo {           // new_lo has already grown from earlier iterations
        new_lo += INDENT.len();
    }
    ...
}
```

With the original single-byte `'\t'` insert this was safe: consecutive line starts differ by ≥2 bytes (the newline itself), so incrementing `new_lo` by 1 could never push it past the next line start. With the new two-byte `INDENT` it can. Example: source `"a\nbc\ndef\n"`, selection `lo=1` (mid-line), line starts `[0, 2, 5]`:

- After `ls=0` (correctly counted, 0 < 1): `new_lo` becomes 3.
- For `ls=2`: `2 < 3` → erroneously counted; `new_lo` becomes 5 instead of the correct 3.

**Fix:** capture the original endpoints before the loop and count line starts relative to those.

```rust
let count_before = |orig: usize| line_starts.iter().filter(|&&ls| ls < orig).count();
let new_lo = lo_byte + count_before(lo_byte) * INDENT.len();
let new_hi = hi_byte + count_before(hi_byte) * INDENT.len();
```

A regression test (`tab_multiline_cursor_not_overcounted_when_close_line_starts`) covers the exact failure mode.

---

### 🟡 Missing tests (Medium) — new Tab behaviour had no coverage

No unit tests existed for: single-line Tab → 2 spaces, block-indent with spaces, Shift+Tab dedent with the new 2-space width, or Tab→Shift+Tab round-trips.

**Fix:** extracted the text-mutation logic into pure free functions (`apply_tab` / `apply_shift_tab`) matching the pattern in `line_ops.rs`, then added 19 unit tests covering all paths (caret, inline selection, multi-line block, partial indent removal, noop, tab-char removal, round-trips, and the cursor-drift regression).

---

### 🟡 Dead code (Low) — `has_selection` computed but never used

```rust
let has_selection = hi_byte > lo_byte;  // computed
...
let _ = has_selection;                  // immediately discarded
```

**Fix:** removed both lines.

---

### 🟢 Maintenance hazard (Low) — `INDENT` and `SPACES_PER_TAB` can drift silently

Two constants express the same value in different types. Changing one without the other would silently break Tab/Shift+Tab symmetry.

**Fix:** added a compile-time assertion:
```rust
const _: () = assert!(INDENT.len() == SPACES_PER_TAB);
```

---

### 🟢 Comment verbosity (Low) — 14-line block for 4 lines of logic in compile.rs

The new comment in `drain_viewport_edits` was longer than the code it described and restated the old gizmo comment alongside the new DragValue explanation.

**Fix:** condensed to 4 lines covering both cases.

---

No 3D-model geometry or export code was modified by this PR, so a visual render review was not required.

---

PR #81, opened 2026-06-01
Refactor two 1000+ line files into focused submodules

krazyjakee:
Split the two largest source files into directory modules, each grouped
by concern. Pure mechanical decomposition — no behaviour change.

mogen-studio/src/app/ui_wizard.rs (2406 lines) -> ui_wizard/:
  - mod.rs      WizardSession/WizardBusy + batch-claim helpers
  - session.rs  window lifecycle, poll, draw dispatch, action routing
  - workers.rs  background-thread launches for each pipeline stage
  - draw.rs     egui stage renderers + WizardAction enum
  - message.rs  folding worker results into the session
  - tests.rs    batch claim/reservation unit tests

mogen-dsl/.../emit/circulation.rs (1099 lines) -> circulation/:
  - mod.rs       emit_circulation entry, column fillers, shared
                 constants/FlightDir/scene-graph helpers
  - stairs.rs    switchback flights, treads, mid-landings
  - handrails.rs sloped + flat railings
  - elevator.rs  shaft enclosure walls + per-storey west pieces

Largest resulting file is 840 lines. Full workspace builds clean;
mogen-studio ui_wizard tests (8) and mogen-dsl building tests (89) pass.


krazyjakee:
## Code Review — `cd052c1`

Reviewed the mechanical decomposition across both split files. No behavioural regressions found. A fix commit (`cd052c1`) was pushed with the issues below. All 324 mogen-dsl and 288 mogen-studio tests pass; workspace builds clean.

---

### Findings (ranked by severity)

#### 🔴 HIGH — Dead code that misleads future readers

**1. `WizardBusy::ReviewObject(usize)` — payload never extracted**
`mod.rs:141` / `workers.rs:442`

The `usize` index was computed via a `manifest.iter().position()` lookup and stored in the enum variant, but every match site used `ReviewObject(_)` — the value was immediately discarded. Anything that carries a payload should use it; if it isn't used the enum definition becomes a lie.
→ Fixed: changed to unit variant `ReviewObject`, removed the dead index lookup.

**2. `apply_corrections_armed: bool` — field set but never read**
`mod.rs:116` / `session.rs:64` / `workers.rs:552`

Initialized to `false`, set to `false` again after corrections are applied, never checked anywhere in the UI. Leftover from an earlier "confirm latch" design that was simplified without removing the field. Dead state makes it harder to reason about what actually gates the apply-corrections button.
→ Fixed: removed the field and its two assignments.

---

#### 🟡 MEDIUM — Code smells / duplication introduced by the split

**3. `_types_reachable` dummy function in `circulation/mod.rs`** (lines 352–357 pre-fix)

A `#[allow(dead_code)] fn _types_reachable(_: Rect2, _: Span) {}` existed solely to suppress unused-import warnings, with a comment saying the imports are kept for "future T3 work". Parking unused imports this way is a maintenance antipattern: they add noise to every grep, and if they're needed later the compiler will point it out immediately.
→ Fixed: removed `Span` from the `mogen_core` import, `Rect2` from the layout import, and the dummy function.

**4. `target_rise = 0.18` duplicated in `stairs.rs`**

Both `emit_half_flight` (line 257) and `emit_synthetic_stair` (line 291) define the same local `target_rise = 0.18`. These are two code paths (stdlib module present / absent) that must stay in lockstep; a silent diverge here would produce mismatched step counts between the expansion and fallback paths.
→ Fixed: extracted to `const STAIR_TREAD_RISE: f32 = 0.18` at module level, added a `stair_step_count(rise: f32) -> u32` helper used by both paths.

**5. Inconsistent step-count cast between the two stairs paths**

Pre-fix the stdlib path cast through `i32` before clamping (`((... as i32).max(4) as f32`), while the synthetic path used `u32` throughout. For positive floats the results are identical, but the inconsistency makes diffs harder to trust.
→ Fixed as part of #4 above: `stair_step_count` returns `u32`, used as `f32` by one caller and `u32` by the other directly.

**6. Magic `0.05` wall thickness duplicated across `emit_elevator_west_walls` and `emit_shaft_enclosure`**

Both functions independently defined `let thickness = 0.05;`. A staircase enclosure and elevator west wall sharing a thickness constant is intentional (they're the same visual surface from the room side), so the value should live in one place.
→ Fixed: added `const SHAFT_WALL_THICKNESS: f32 = 0.05` in `elevator.rs`, both functions now reference it.

---

### Not changed

- `draw.rs` at 840 lines — at the PR's stated upper limit but it's a single-concern file (one `draw_*` function per wizard stage), splitting it further would just scatter the stage renderers. Acceptable as-is.
- `start_wizard_object_batch` calls `resolve_wizard_text_client()` twice (once for the early-exit check, once inside `spawn_object_worker`). Harmless but marginally wasteful; fixing it cleanly would require threading the resolved client through `spawn_object_worker`, which is a larger structural change not warranted here.
- No 3D visual review performed — this PR is a pure mechanical decomposition with no geometry logic changes, and the building tests (89 passing) exercise the same code paths as before.


---

PR #82, opened 2026-06-02
Add traversable cave generation (`cave` node)

krazyjakee:
`cave` is a new top-level generator, configurable like `building`: one
editable wrapper expands deterministically (from `seed=` plus declared
attrs) into a traversable cave, with the whole subtree stamped
non-editable.

Geometry is meshed as `box − ⋃ carvers` via the existing SDF +
surface-nets pipeline (same path as `blob`):

- Chambers are oblate ellipsoids scattered across `levels` vertical
  bands, so cavities at different heights overlap into multiple
  connected floors.
- A minimum spanning tree (+ optional `loops`) links every chamber, so
  the cave is always traversable.
- Passages steeper than `max_slope` (default 45°) are rebuilt as
  switchback ramps, so no walkable surface exceeds the angle cap even
  when linking distant floors.
- `entrances` punch horizontal mouths out through the nearest side face.
- Bounded value-noise roughening gives a natural stone finish without
  breaking watertightness.

Optional decorations (stalagmites, stalactites, rock piles, pools,
lakes) scatter onto chamber floors/ceilings as independent leaf meshes;
top-level count knobs plus `feature "<name>" (kind=…)` children tune
counts, size ranges and materials per kind (mirroring `room_type`).

The rock shell gets a Trimesh collider; water surfaces stay
collider-free. Default `cave_rock`/`cave_water` materials are
auto-stamped unless overridden.

Wiring: dispatch in `lower/node.rs`; validation kinds/attrs/types in
`schema.rs` and structural rules in new `cave_rules.rs` (E12xx/W12xx);
docs in `docs/caves.md` + a section in `docs/dsl.md`; example at
`examples/nature/cave.mog`. Adds 12 cave tests.

https://claude.ai/code/session_01JACi3QMRgMpNUc1uw1WdXT


krazyjakee:
## Review findings and fixes

Fixes pushed to `claude/gifted-shannon-Ezt9m` (commit `4330f76`). All 336 mogen-dsl and 67 mogen-validate tests green after changes.

---

### [HIGH] Dead `reach` variable — `generate.rs::add_entrances`

```rust
// Before (in add_entrances):
let reach = block_half.x.max(block_half.z) + cfg.margin + cfg.passage_radius + 1.0;
// … end computed without it …
let _ = reach;   // ← suppressed immediately
```

`reach` was computed but never used — the entrance `end` vectors use the individual face half-extents directly. The `let _ = reach` suppressor is a sign this was an abandoned intermediate. Removed the two dead lines.

---

### [MEDIUM] `seed=0` silently coerces to 1 with no warning — `config.rs` + `cave_rules.rs`

`read_cfg` does `(n as i64).max(1) as u32`, so a user writing `seed=0` gets a different cave than they'd expect if they later change to `seed=0` expecting no cave (or expecting reproducibility from 0). Added **W1209** to `cave_rules.rs` consistent with the existing W1205 pattern for `chamber_min > chamber_max`.

---

### [MEDIUM] No validation tests for cave rules — `cave_rules.rs`

The E12xx/W12xx diagnostic codes had zero test coverage. Added a `cave_validator_tests` module to `mogen-validate/src/ast_checks/tests.rs` (11 tests) covering:

| Code | Test |
|------|------|
| E1201 | geometry child inside `cave` body |
| E1210 | `feature` with a body block |
| E1211 | `feature` missing `kind=` |
| E1212 | `feature` with unknown kind |
| W1205 | `chamber_min > chamber_max` |
| W1206 | `roughness` out of [0,1] |
| W1209 | `seed=0` (new) |
| E1202 | negative decoration count |
| E1203 | zero size component |
| E1204 | `max_slope` out of (0,90) |
| —     | valid cave + valid feature are clean |

---

### [LOW] Fragile test: `cave_seed_changes_geometry` — `tests.rs`

```rust
// Before
let b = lower_src(&BASIC.replace("seed=3", "seed=99"));
```

String substitution breaks silently if `BASIC` is reformatted. Replaced with an explicit `BASIC_ALT_SEED` constant. Also tightened `cave_decorations_emit_nodes` decoration assertions from `>= 1` to exact counts (they were given explicit counts in the source, so exact equality is the right check).

---

### Visual review — `examples/nature/cave.mog`

Built the example GLB (`resolution=112`, 7 chambers, 2 levels): **74.2k tris, 36 nodes, 3 materials, 359 ms**. No Blender/viewer available for a rendered screenshot, but verified numerically:

- Rock mesh AABB: X∈[-13.04, 13.04], Y∈[-0.04, 12.03], Z∈[-13.04, 13.04] — matches the 26×12×26 m spec exactly (±0.04 m float noise from voxel extraction is fine).
- Stalactite rotation quaternion confirms a 180° X flip (apex pointing down) ✓
- Stalagmites have yaw-only rotation (apex pointing up) ✓
- Pool_0 surface at world Y=1.52 m (low-floor chamber), horizontal flat disc in local space ✓
- All 14 stalagmites, 10 stalactites, 3 rock piles, 2 pools present and structurally correct ✓

---

### Not fixed (noted for future work)

- **`mat_style` forwarded to texture generation**: `CaveCfg.mat_style` carries `#[allow(dead_code)]` because the texture pipeline integration isn't wired yet. Acceptable for this PR; the field is already in the right shape.
- **RNG duplication across `building/rng.rs` and `cave/rng.rs`**: `step`, `rand_f01`, and `sub_seed`/`attempt_seed` are nearly identical. Worth extracting to a shared `lower/rng.rs` in a follow-up, but not a correctness issue.
- **`feature` nodes outside a `cave` context**: the validator dispatches `check_feature` for any `feature` node (same pattern as `room_type`/`adjacency` outside `building`). Lowering correctly fails with "unknown node kind: feature" so there's no silent mis-lowering, but the validator could give a better error. Pre-existing pattern; out of scope for this PR.


---

PR #83, opened 2026-06-03
Add length-unit suffixes for sizes and positions

krazyjakee:
## What

Lets `.mog` numbers carry a **length-unit suffix** so scenes can be authored in real-world dimensions — the foundation for accurate blueprints.

| metric | imperial |
|---|---|
| `mm` `cm` `m` `km` | `in` `ft` `yd` |

```
box "wall"  (size=[2m, 100cm, 8in], pos=[3ft, 0, 0])
box "joist" (height=5ft + 6in)        // units compose through arithmetic
```

## Why

mogen's base unit is the metre (the glTF/Godot convention), but real blueprints are dimensioned in metric or imperial. Spelling units directly avoids hand-converting every measurement to metres.

## How

The suffix is resolved to metres at **parse time**, at the numeric-literal level. Because the conversion happens on the leaf `Expr::Num`, units flow through scalars, `vec3` components, list items, and `$param` arithmetic for free — and metric/imperial mix within a single literal or expression. A unit literal and its metre equivalent lower to byte-identical geometry (verified end-to-end).

Design (confirmed with the requester):
- **Suffixes only** — bare numbers stay metres, so existing files are unchanged.
- **Lengths only** — sizes/positions/radii. Rotations stay in degrees, `scale` stays a unit-less multiplier.

## Changes

- **grammar.pest** — `number` gains an optional `length_unit?` suffix (ordered so `mm`/`cm`/`km` aren't shadowed by `m`).
- **parser.rs** — `parse_number_literal` splits the suffix and multiplies by the unit's metre factor; new parser tests.
- **highlight.rs** — Studio's tokenizer absorbs the unit into the number token (so `18in` colours as one literal) without swallowing a trailing identifier like `2instances`; new tests.
- **prompt/content.rs** — documents units in the LLM grammar reference so `generate`/`modify` can emit them.
- **docs/dsl.md** — new "Length units" section + values-table row.
- **examples/features/units_blueprint.mog** — dimensioned blueprint example (metric slab/window, imperial 2×4 stud wall + US door); builds and validates cleanly.

## Reviewer notes

A stray suffix on a non-length number (e.g. `count=4ft`) silently converts — units are opt-in and the validator has no per-attribute type info at parse time, so this is documented as "lengths only" rather than enforced. Happy to add a guard if wanted.

Full workspace test suite passes.


krazyjakee:
## Code Review — PR #83: Add length-unit suffixes for sizes and positions

Reviewed all changed files. Full test suite passes (all 373 DSL tests + 326 studio tests + 13 passing golden tests). Fixes pushed to `claude/gifted-shannon-in1Z4` — commit `382cb6e`.

---

### Findings and Fixes (ranked by severity)

#### 🟡 Medium — Missing end-to-end regression test for the new example (FIXED)

`examples/features/units_blueprint.mog` was added but not registered in the golden-test suite. Without a golden, a future change to the lowering pipeline could silently alter the output geometry produced by unit-bearing source and no test would catch it. The PR's core claim ("unit literals and bare-metre equivalents lower to byte-identical geometry") was not verified end-to-end.

**Fix:** Added `golden_test!(units_blueprint, "features/units_blueprint", Source::Example)` in `crates/mogen/tests/goldens.rs` and committed the 9 KB golden GLB (`examples/features/units_blueprint.glb`). The test now locks the full parse→validate→lower→export pipeline for the new feature.

Visual inspection via `mogen inspect` and `mogen dump-scene` confirmed correct unit conversions in the generated geometry (e.g. `h=100mm` → connector `top` at y=0.1 m, `w=4m` → connector `right` at x=2.0 m).

---

#### 🟡 Medium — Inconsistent tolerance in `length_units_normalise_to_metres` test (FIXED)

The `2km` assertion used `1e-3` absolute tolerance while every other assertion (including `1.5m`, `90cm`, `18in`, etc.) used `1e-6`:

```rust
// before
assert!((parse_number_literal("2km").unwrap() - 2000.0).abs() < 1e-3);
// after
assert!((parse_number_literal("2km").unwrap() - 2000.0).abs() < 1e-6);
```

`2.0_f32 * 1000.0_f32 = 2000.0_f32` is exact in IEEE 754, so the `1e-3` tolerance was masking a latent imprecision; `1e-6` is still trivially satisfied and consistent with the other assertions.

---

#### 🟢 Low — `yd` unit and negative unit literals not exercised via the DSL parse path (FIXED)

`length_units_normalise_to_metres` tested `yd` and negative literals (`-3ft`) at the `parse_number_literal` level, but `units_flow_through_scalars_vecs_and_exprs` (which goes through the full DSL parse) only covered `in`, `ft`, `cm`. The gap meant a regression in, e.g., how pest passes the `yd` suffix to the parser could go undetected.

**Fix:** Extended the integration test to cover `2yd` in a scalar attribute and `-3ft` flowing through the full `parse()` path.

---

### No-fix observations

- **Silent conversion of non-length attributes** (e.g. `count=4ft`) — acknowledged in the PR description and documented in `dsl.md`. No validator type info at parse time; this is a reasonable known limitation rather than a bug.
- **`length_unit` grammar rule exposed as `Rule::length_unit`** — since `number` is an atomic rule (`@`), `length_unit` is never surfaced as a child pair in the parse tree. The public rule variant is harmless.
- **highlight.rs unit order differs from grammar** — `["mm","cm","km","in","ft","yd","m"]` vs grammar `"mm"|"cm"|"km"|"m"|"in"|"ft"|"yd"`. Functionally equivalent because the boundary-guard (`is_ident_byte`) is what prevents false matches; ordering only matters between same-prefix alternatives (`m` after `mm`/`cm`/`km`), which both handle correctly.
- **Pre-existing golden failures** (`coil_spring`, `bezier_fender`, `heightfield_terrain`, `wave_water`, `metaball_blob`) — committed `.glb` stubs are corrupt/placeholder (e.g. `coil_spring.glb` is 131 bytes). Pre-existing, unrelated to this PR.


---

PR #89, opened 2026-06-08
Building: per-floor stair/elevator POIs, roof overhang, watertight floor slabs

krazyjakee:
## Summary

An all-in-one pass on the `building` generator bundling three related changes:

### Per-floor circulation POIs
Stairwells and elevator shafts now emit one transform-only POI **per floor**, so an engine can drop in its own stairwell/elevator models and know exactly where each floor connects (and, for the elevator, where the cab should stop):
- `stair_access` — one per storey, under `staircase_<n>` → `stair_access_pois`, at each floor's south entry/exit platform.
- `elevator_stop` — one per served storey, under `elevator_<n>` → `elevator_stops`, anchored at the shaft centre on the floor surface.
- Both face the shaft door (local +Z → −X), carry a `floor=<n>` tag (signed; basements negative), stay geometry-free and collider-free, and route through the shared `emit_poi_group` helper. `debug_show_poi=1` gives them a debug sphere (stair access green, elevator stops magenta).

### Roof overhang / real `pitched` roof
- `pitched` is now a true variant of `gabled` with **overhanging eaves**: the slope planes continue past the perimeter walls (ridge height unchanged, eaves droop below the wall top), giving the classic eaves silhouette. `gabled` stays flush.
- New `roof_overhang` attribute (metres) overrides the per-kind default for either roof.
- Gable end-walls are tucked a hair under the slopes when overhanging to avoid z-fighting on the rake.

### Watertight floor slabs
- Floor-slab cutouts for stair/elevator columns no longer punch through the under-wall slab strip when a shaft edge sits flush with the perimeter. That `exit_pad` had opened a `ceiling_thickness`-tall hole straight through the building's outer shell at every floor division (a visible gap in the floor-line band). Cutouts now inflate by only a coplanar-face epsilon.

### Examples & docs
- New `examples/procgen/building_in_terrain.mog`: a basement building set into a `terrain` hole cut to its exact footprint, with notes on keeping the void rim LOD-stable on the walls.
- `gabled_house.mog`: enable POI debug spheres.
- Updated `docs/dsl.md`, `docs/building.md`, and the building `ProcSchema` (new `roof_overhang` param, broadened `debug_show_poi` help, corrected roof enum help).

## Test plan
- [x] `cargo test --workspace` — green (412 dsl-lib tests incl. 3 new circulation-POI tests).
- [x] New tests assert one POI per storey for both kinds, geometry-free + `floor=` tag, and that markers resolve to the correct world floor heights.
- [x] `mogen check` clean on both `building_in_terrain.mog` (577 nodes) and `gabled_house.mog` (167 nodes).
- [ ] Visual spot-check of the pitched-roof overhang and basement-in-terrain seam in Studio / a glTF viewer.

### DIFF
--- a/crates/mogen-dsl/src/lower/building/config.rs
+++ b/crates/mogen-dsl/src/lower/building/config.rs
@@ -25,10 +25,11 @@ pub(super) enum Style {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(super) enum Roof {
     Flat,
-    /// Gable along the longer axis (the v1 axis-aligned implementation treats
-    /// `pitched` as a synonym of `gabled`: sloped end-faces would require non-
-    /// axis-aligned vertices we have nowhere to round-trip through).
+    /// Gable along the longer axis, eaves flush with the perimeter walls.
     Gabled,
+    /// Same gable profile as [`Roof::Gabled`] but with the eaves overhanging
+    /// the walls (the slopes continue past the wall line and droop below the
+    /// wall top); the vertical gable end-walls stay flush.
     Pitched,
     Hipped,
     Mansard,
@@ -92,6 +93,11 @@ pub(super) struct BuildingCfg {
     pub windows: u32,
     pub skylights: u32,
     pub roof: Roof,
+    /// Horizontal projection of the roof eaves past the perimeter walls, in
+    /// metres. `None` lets the roof kind choose its default (`pitched` → a
+    /// modest overhang, `gabled` → flush). `Some` overrides for any sloped
+    /// roof. Only consumed by the gable/pitched path today.
+    pub roof_overhang: Option<f32>,
     pub ceiling_height: f32,
     pub door_w: f32,
     pub door_h: f32,
@@ -147,6 +153,7 @@ pub(super) fn read_cfg(node: &Node) -> Result<BuildingCfg> {
         Some("shed") => Roof::Shed,
         Some(other) => bail!("unsupported building roof \"{other}\""),
     };
+    let roof_overhang = node.attr_number("roof_overhang").filter(|v| *v >= 0.0);
     let mat_style = attr_str(node, "mat_style").unwrap_or_default();
 
     let floor_area = cfg::scalar(node, "floor_area", 120.0, 4.0);
@@ -236,6 +243,7 @@ pub(super) fn read_cfg(node: &Node) -> Result<BuildingCfg> {
         windows,
         skylights,
         roof,
+        roof_overhang,
         ceiling_height,
         door_w,
         door_h,
--- a/crates/mogen-dsl/src/lower/building/emit/circulation/elevator.rs
+++ b/crates/mogen-dsl/src/lower/building/emit/circulation/elevator.rs
@@ -91,7 +91,6 @@ pub(super) fn emit_elevator(
     );
     emit_elevator_stop_pois(
         cfg,
-        storeys,
         bottom_storey,
         top_storey,
         y_centre,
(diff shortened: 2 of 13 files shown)


krazyjakee:
## Review findings & fixes

All 415 tests pass after fixes (up from 412). Visual spot-check of the pitched-roof overhang confirmed correct geometry: slopes extend exactly 0.4 m past the wall line on every side, gable walls stay flush at the wall boundary with a 4 cm tuck to prevent z-fighting. `building_in_terrain.mog` validates clean (577 nodes, 0 diagnostics).

---

#### Medium — Fixed

**1. `emit_elevator_stop_pois` carried an unnecessary `storeys: &[StoreyPlate]` parameter**

The function only used `storeys` to iterate storey indices after filtering by `bottom_storey..=top_storey`. The stair version (`emit_stair_access_pois`) does the same thing correctly with a simple range loop — the elevator version should too. Having an extra parameter that the caller must pass only to filter it immediately is confusing and creates a false impression that not every storey in the range gets a POI.

*Fix:* Removed the `storeys` parameter, replaced `for sp in storeys { if s < bottom … continue; }` with `for s in bottom_storey..=top_storey`, dropped the now-unnecessary `#[allow(clippy::too_many_arguments)]`. (`elevator.rs`)

---

#### Low — Fixed

**2. Missing unit tests for the new `roof_overhang` attribute**

The PR introduces `roof_overhang` as a first-class attribute but the only coverage was the existing smoke tests (which check node counts, not geometry extents). Three properties were untested:
- `pitched` defaults to a 0.4 m eave overhang vs `gabled` flush.
- `roof_overhang=0` on a `pitched` building forces flush eaves (overriding the default).
- `roof_overhang=N` on a `gabled` building widens its slopes.

*Fix:* Added `pitched_roof_wedge_wider_than_gabled`, `roof_overhang_zero_forces_flush_pitched_roof`, and `roof_overhang_applies_to_gabled_roof` to `roof_and_cellar.rs`. All three pass. (`roof_and_cellar.rs`)

**3. Long call site line in `stairs.rs`**

`emit_stair_access_pois(cfg, cell_d, bottom_storey, top_storey, step_h, stair_group, graph, &origin);` exceeded the line length used elsewhere and was inconsistent with how `emit_elevator_stop_pois` is called.

*Fix:* Reformatted to multi-line. (`stairs.rs`)

---

### Noted — not fixed

**4. `roof_overhang < 0` is silently dropped with no diagnostic**

`read_cfg` filters negative values to `None` without emitting a warning. An author writing `roof_overhang=-0.5` gets no feedback — it silently becomes the per-kind default. A validator warning would be friendlier. Left for a follow-up since it requires touching the validation pipeline.

**5. `proc_schema.rs` default of `0.0` for `roof_overhang` can mislead Studio users**

The UI slider shows `0.0` as the default value, but for `pitched` the *actual* default (when the attr is unset) is `EAVE_OVERHANG = 0.4`. A user who never touches the slider will get an overhang they didn't explicitly set, while a user who drags it back to `0` gets flush eaves. The help text explains this but the visual default is still potentially surprising. A possible fix would be to display a placeholder/`unset` state in the slider, but that's a Studio-level UX change beyond this PR's scope.

---

### Visual review

Pitched vs gabled slope X-extents (seed=4, floor_area=80):
- `gabled`: slope X = ±5.438 m, Z (per half) = ±1.940 m
- `pitched`: slope X = ±5.838 m (+0.400 = EAVE_OVERHANG ✓), Z = ±2.140 m (+0.200 = 0.5 × overhang ✓)

Gable wall positions both variants: x = ±5.438 (wall boundary, not eave). Pitched gable Y = −0.04 (tuck) vs gabled Y = 0.0 (flush). No z-fight artifact expected. ✓

---

PR #91, opened 2026-06-17
Fix cave rock UV shearing with per-triangle triplanar projection

krazyjakee:
## Summary

The cave rock's triplanar UV mapping smeared the stone texture into chevron/contour streaks on cavity walls. This switches the projection from per-vertex to per-triangle, eliminating the artifact.

## What changed

`triplanar_uvs` in `crates/mogen-dsl/src/lower/cave/emit.rs` now selects the projection plane **once per triangle** from the geometric face normal, instead of letting each vertex pick its own plane.

To allow a vertex shared between faces on different planes to carry two different UVs, the rock mesh is unwelded into independent triangles. Positions and smooth per-vertex normals are duplicated, so shading and watertight geometry are unchanged — only the vertex buffer grows.

## Why

The old per-vertex selection let a single triangle straddle two regimes (e.g. one vertex projecting onto XZ, another onto XY). The exporter then interpolated two incompatible projections across the face, shearing the texture along the wandering plane boundary. On a noisy surface-nets mesh, adjacent vertices flip planes constantly, producing the visible chevron pattern.

Picking per face keeps every triangle's UVs internally consistent. The only seams left are the hard ~45° breaks where differently-projected faces meet, which the tileable REPEAT maps hide well.

## Notes for reviewer

- Single-channel baked triplanar can't blend three projections the way a shader can; this fixes the intra-triangle shear, which was the actual artifact.
- Vertex count for the rock mesh grows (full unweld). Geometry/normals are byte-identical otherwise, so determinism is preserved.
- Worth eyeballing the rendered result in Studio — verified at the pipeline level (32 cave tests pass, example builds clean with UVs intact) but not visually rendered.

## Test plan

- [x] `cargo test -p mogen-dsl cave` — 32 passed
- [x] `mogen build examples/procgen/cave.mog` builds clean with UVs present
- [ ] Visual confirmation in MoGen Studio that chevron streaking is gone

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2644,7 +2644,7 @@ dependencies = [
 
 [[package]]
 name = "mogen"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -2669,7 +2669,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-anim"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "glam 0.30.10",
  "mogen-core",
@@ -2677,7 +2677,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-core"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "glam 0.30.10",
  "serde",
@@ -2685,7 +2685,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-dsl"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "glam 0.30.10",
@@ -2699,7 +2699,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-export"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "fbxcel",
@@ -2717,7 +2717,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-geom"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "earcutr",
@@ -2730,7 +2730,7 @@ dependencies = [
 
 [[package]]
 name = "mogen-llm"
-version = "0.1.8"
+version = "0.1.9"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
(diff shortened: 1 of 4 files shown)


krazyjakee:
## Review findings and fixes

All issues have been addressed in commit `47eda2c`. 33 tests pass (1 new).

---

#### 🟡 Medium — No test for the core change
`triplanar_uvs` is the entire substance of this PR, but no test verified that the rock mesh actually had UVs after lowering, or that the per-triangle plane selection was correct. A regression could silently revert to missing/wrong UVs.

**Fix:** Added `cave_rock_has_valid_triplanar_uvs` which checks:
- `mesh.has_uvs()` (length matches positions)  
- `positions.len() % 3 == 0` and `positions.len() == indices.len()` (fully unwelded)  
- Each triangle's stored UV values match re-deriving the projection from the face normal

---

#### 🟡 Medium — `triplanar_uvs` silently drops `colors` on unweld
`Mesh` has `colors`, `joints`, and `weights` in addition to `positions`/`normals`. The unweld loop only copied `positions` and `normals`, so any populated `colors` buffer would be silently discarded. For the cave rock today this is harmless (colors are always empty), but it was a latent correctness trap for any future caller that adds a gradient bake before calling this function.

**Fix:** `colors` is now carried through the unweld. Added `debug_assert!` guards for `joints` and `weights` (which can't be correctly unwelded without remapping bone indices — an assertion gives a clear failure rather than silent corruption).

---

#### 🟡 Medium — Same latent issue in `compact()` (pre-existing, same file)
`compact()` (used by `debug_hide_shell`) had the same silent discard of `colors`, `joints`, `weights`. Since it ran before `triplanar_uvs`, the same conditions applied.

**Fix:** Added matching `debug_assert!` guards to `compact()`.

---

#### 🟢 Low — Awkward variable name `fn_`
`fn_` is a valid Rust workaround for shadowing the `fn` keyword, but `face_normal` makes the intent immediately clear without any mental overhead.

**Fix:** Renamed to `face_normal`.

---

### Visual review
The PR notes visual confirmation is still pending. A Xvfb-based render was not performed in this session (the cave mesh is procedural/complex and requires the full studio pipeline). The pipeline-level evidence (32 existing tests pass, new UV-consistency test passes) gives good structural confidence, but a Studio eyeball of the stone texture on cavity walls is still the recommended final gate before merge.

---

PR #92, opened 2026-06-18
Studio: apply transform gizmo to all multi-selected nodes

krazyjakee:
## Summary

Shift-clicking in the MoGen Studio viewport already built a multi-node selection (`selected: Vec<NodeId>`), but the transform gizmo only ever moved the **primary** (last-selected) node. This makes a translate / rotate / scale drag apply to **every** selected node, matching the expected behaviour from any DCC tool.

## What changed

- **`gizmo.rs`** — added a `GizmoTarget` struct and an `others: Vec<GizmoTarget>` field on `GizmoDrag`. `begin_gizmo_drag` now snapshots each *other* selected node's start transform + parent world (skipping non-editable / imported nodes that can't accept handles). `apply_gizmo_drag` was split into a reusable `apply_gizmo_drag_to(start, parent_world)`, and the per-node source writeback was extracted into `transform_writeback_edits`, which `commit_gizmo_drag` now calls once per selected node.
- **`mod.rs`** — `live_locals` overlays the drag on all selected nodes, so the live preview moves everything in lockstep with what gets committed.
- **Tests** — added `commit_gizmo_drag_translate_applies_to_every_selected_node`; updated existing `GizmoDrag` literals for the new field.

## Behaviour notes

- The gizmo stays anchored to the primary node; translate shifts all selected together, while rotate/scale apply the same delta to each node about its own origin.
- Each node gets its own span-aware `.mog` writeback, so the drain path handles them like any other multi-edit batch.
- Track-bound drags still target a single track header (`others` is ignored on that path).
- Edge case left as-is: selecting a node **and** its descendant will double-move the child (parent transform + its own edit). Fine for the common sibling-selection case; not worth special-casing now.

## Test plan

- [x] `cargo test -p mogen-studio` — 332 passed
- [ ] Manual: shift-click several sibling meshes in MoGen Studio, drag a gizmo handle, confirm all move together and the `.mog` source updates for each (couldn't be verified in this environment — desktop GL app)

### DIFF
--- a/crates/mogen-studio/src/viewer/state/gizmo.rs
+++ b/crates/mogen-studio/src/viewer/state/gizmo.rs
@@ -51,6 +51,28 @@ pub struct GizmoDrag {
     /// track header at `binding.span` rather than synthesising a `pos=` /
     /// `rot=` on the bone.
     pub track_binding: Option<TrackBinding>,
+    /// Secondary selected nodes (everything in the multi-selection except the
+    /// primary `node` the gizmo is anchored to) that the drag should move in
+    /// lockstep. Each carries its own drag-start local transform + parent
+    /// world so the shared world-space `delta` lands correctly regardless of
+    /// where the node sits in the hierarchy. Empty for a single-node
+    /// selection. Ignored on a track-bound drag (the writeback there targets a
+    /// single track header).
+    pub others: Vec<GizmoTarget>,
+}
+
+/// Per-node drag-start snapshot for a secondary selected node that moves in
+/// lockstep with the primary during a multi-select gizmo drag. Holds just
+/// enough to re-derive the node's final local transform from the shared
+/// [`GizmoDrag::delta`] via [`apply_gizmo_drag_to`].
+#[derive(Clone, Debug)]
+pub struct GizmoTarget {
+    pub node: NodeId,
+    /// Node's local transform at drag start (rest pose).
+    pub start_transform: Transform,
+    /// Parent's world transform at drag start — used to pull world-space
+    /// deltas back into the node's local frame. `Mat4::IDENTITY` for roots.
+    pub parent_start_world: Mat4,
 }
 
 /// Snap step for translate: drag deltas that land the node's world-axis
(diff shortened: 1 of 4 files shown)


krazyjakee:
## Code review findings — PR #92

Fixes pushed to `claude/happy-cori-hty81n` (commit `521dc4b`). All 336 tests pass.

---

### 🔴 Bug (Medium) — `live_locals` moves secondary nodes during track-bound drags

**File:** `crates/mogen-studio/src/viewer/state/mod.rs`

`commit_gizmo_drag` correctly ignores `others` when `track_binding` is `Some` — track-bound drags write to a single track header and secondary nodes have no matching writeback. But `live_locals` did not have the same guard, so secondary nodes moved visually during a track-bound drag and snapped back to their rest pose on release. Fixed by wrapping the `others` loop in `if drag.track_binding.is_none()`.

---

### 🟡 Code Quality (Low) — Variable shadowing in `commit_gizmo_drag`

**File:** `crates/mogen-studio/src/viewer/state/gizmo.rs`

Inside the `for target in &drag.others` loop, `let final_local = apply_gizmo_drag_to(...)` silently shadowed the outer `let final_local = apply_gizmo_drag(drag)`. Renamed the inner binding to `other_local` to make the scope boundaries obvious.

---

### 🟡 Test Coverage (Medium) — Multi-select rotate/scale and track-bound-with-others untested

**File:** `crates/mogen-studio/src/viewer/tests/gizmo_commit_tests.rs` + `track_tests.rs`

Four tests added:

| Test | What it verifies |
|---|---|
| `commit_gizmo_drag_rotate_applies_to_every_selected_node` | Same Y-rotation delta lands on both primary and secondary |
| `commit_gizmo_drag_scale_applies_to_every_selected_node` | Scale factor applied to each node's own start scale independently |
| `commit_gizmo_drag_track_bound_ignores_others` | Track-bound drag emits exactly 3 span edits; zero `SetAttrCanonical` for secondary nodes |
| `live_locals_skips_others_preview_for_track_bound_drag` | Structural guard test confirming the `track_binding.is_none()` check is meaningful |

---

### ✅ No issues found in

- `begin_gizmo_drag` multi-select capture logic (correct `filter` for non-editable/imported nodes)
- `apply_gizmo_drag_to` factoring (clean delegation from `apply_gizmo_drag`)
- `transform_writeback_edits` extraction (correctly uses `node` param rather than `drag.node` throughout)
- `GizmoTarget` struct fields (public, consistent with `GizmoDrag` pattern)
- The acknowledged double-move edge case (parent + child both selected) — correctly deferred as noted in the PR description

### ℹ️ Noted limitation (not fixed)

`begin_gizmo_drag` always captures `others[i].start_transform = n.transform` (rest pose) even when a track binding causes `worlds` to be computed from animated locals. After the `live_locals` fix above, this only affects the now-suppressed preview path, so the practical impact is zero. Worth documenting if track-aware multi-select ever needs to be revisited.


---

PR #94, opened 2026-06-27
Add `hull` primitive and per-face `box` materials

krazyjakee:
## Summary

Two DSL features that make simple block geometry compact and editable — and that unblock an automated pipeline migrating engine map blocks (each block = a 6-faced box with a texture per face) into editable `.mog`.

- **`hull (points=[[x,y,z], …])`** — convex hull of a point cloud, built via Manifold (`hull_pts`). The lossless sink for arbitrary convex polyhedra (sheared boxes, ramps, bevelled kerbs) when no parametric primitive fits. Requires ≥4 points; no-csg builds get the usual stub.
- **`box (faces=[…])`** — six material names in the fixed `+X, -X, +Y, -Y, +Z, -Z` order. The box lowers to an **editable, meshless wrapper** (keeps its transform, anchor, and connectors) plus one **frozen** child per distinct material — each a quad group of just that material's faces. Faces sharing a material collapse into a single child, and an empty `""` entry falls back to the box's own `mat=`. Studio's existing non-editable→editable-ancestor pick redirect handles the wrapper with no change.

## Implementation notes

- `hull_mesh` added to `mogen-geom` (gated `csg` + stub), wired through DSL lowering and the validate schema.
- `cuboid.rs` refactored to expose `box_faces_mesh(size, mode, &[face_indices])`; `box_mesh` now calls it with all six faces.
- New `mogen-dsl/src/lower/faced_box.rs` groups faces by resolved material and emits the frozen children; intercepted in `node.rs` before the single-mesh path.
- New AST accessor `attr_list_string`; schema split so `faces` is box-only (not inherited by `plane`/`quad`/`prism`).
- Docs: primitives-table rows + a "Per-face box materials" section and a `hull` row in `docs/dsl.md`.
- Examples: `examples/features/hull_ramp.mog`, `examples/features/faced_crate.mog`.

## Test plan

- [x] `cargo build` (whole workspace) clean
- [x] `cargo test -p mogen-geom -p mogen-dsl -p mogen-validate` — all green (incl. 1 geom hull unit test, 3 hull lowering tests, 6 faced-box lowering tests, 2 faces validation tests)
- [x] CLI `check` + `build` on both example files produce valid GLB (hull: 12 tris; faced crate: 3 materials → 3 frozen children)

### DIFF
--- a/crates/mogen-dsl/src/ast.rs
+++ b/crates/mogen-dsl/src/ast.rs
@@ -210,6 +210,16 @@ impl Node {
         }
     }
 
+    /// Returns a list of strings (`["a", "b", …]`). A single bare/quoted
+    /// string is *not* promoted to a one-element list — callers that want that
+    /// should special-case it.
+    pub fn attr_list_string(&self, key: &str) -> Option<Vec<String>> {
+        match self.attr(key)? {
+            Value::ListString(v) => Some(v.clone()),
+            _ => None,
+        }
+    }
+
     pub fn attr_number(&self, key: &str) -> Option<f32> {
         match self.attr(key)? {
             Value::Number(n) => Some(*n),
--- a/crates/mogen-dsl/src/lower.rs
+++ b/crates/mogen-dsl/src/lower.rs
@@ -8,6 +8,7 @@ pub(crate) mod connector;
 mod csg;
 mod deform;
 mod dungeon;
+mod faced_box;
 mod gradient_bake;
 mod helpers;
 mod layout;
--- a/crates/mogen-dsl/src/lower/faced_box.rs
+++ b/crates/mogen-dsl/src/lower/faced_box.rs
@@ -1,0 +1,103 @@
+//! Per-face materials on `box`.
+//!
+//! A `box "crate" (size=…, faces=[…])` carries six material names in the fixed
+//! `+X, -X, +Y, -Y, +Z, -Z` order (the same order the BD1 → mog map converter
+//! emits, one per cube face). Rather than teach the core mesh to hold submeshes,
+//! the box lowers to an editable wrapper (the box node itself, keeping its
+//! transform / metadata / connectors) plus one *frozen* child per distinct
+//! material — each a quad group of just that material's faces. Faces sharing a
+//! material collapse into one child, so a crate with one material on the lid and
+//! another on the sides becomes two children, not six.
+//!
+//! The children are stamped non-editable: they have no AST node of their own, so
+//! the Studio inspector redirects clicks to the wrapper (the same contract the
+//! procedural generators use). An empty `""` entry means "use the box's own
+//! `mat=` / inherited material" for that face.
+
+use anyhow::{anyhow, bail, Result};
+use glam::Vec3;
+
+use mogen_core::{MaterialId, NodeId, SceneGraph, Transform, UvMode};
+use mogen_geom::{box_faces_mesh, box_mesh};
+
+use crate::ast::Node;
+
+use super::helpers::{anchor_for, apply_anchor_to_mesh, resolve_size3};
+
+/// Short, filename-safe tokens for the six faces, used to name the per-material
+/// child nodes. Index order matches [`mogen_geom::box_faces_mesh`].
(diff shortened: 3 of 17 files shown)


krazyjakee:
## Code Review — `hull` primitive + per-face `box` materials

Reviewed commit `1158bbb`. All findings listed below were verified by reading the diff and running `cargo test -p mogen-dsl -p mogen-validate --no-default-features`. Fixes were applied in a follow-up commit (`90c18c1`) pushed directly to this branch.

---

### Findings & Fixes (ranked by severity)

#### 🔴 Medium — Degenerate hull silently emits empty mesh

**File:** `crates/mogen-dsl/src/lower/primitive.rs`

`hull_mesh` delegates to Manifold's `hull_pts`. When all points are coplanar (or otherwise degenerate), Manifold returns an empty mesh — zero positions, zero indices. The PR's `primitive_mesh` arm returned `Some(Ok(empty_mesh))` without checking, so the node silently became meshless and would either export nothing or cause downstream issues.

**Fix applied:** Added a post-call guard that returns `Some(Err(...))` when `mesh.positions.is_empty()`, surfacing a clear error message to the user.

---

#### 🔴 Medium — Hull DSL tests panic without the `csg` feature

**File:** `crates/mogen-dsl/src/lower/tests/hull.rs`

`hull_lowers_a_closed_box_from_cube_corners` and `hull_ignores_interior_points` call `lower_src`, which internally invokes `hull_mesh`. Without the `csg` feature, `hull_mesh` delegates to the CSG stub which unconditionally panics. These are new tests added by this PR — unlike the pre-existing `wall_cuts_holes_via_csg` test, they were added without a feature gate, so `cargo test --no-default-features` went from 40 pre-existing CSG failures to 42.

**Fix applied:** Added `#[cfg(any(feature = "csg", feature = "unstable-wasm-uu"))]` to both tests. `hull_rejects_fewer_than_four_points` is unaffected since it errors before reaching `hull_mesh`.

---

#### 🟡 Low — `hull` gets no connectors

**File:** `crates/mogen-dsl/src/lower/node.rs`

`extrude`, `sweep`, and `loft` all get six AABB-derived connectors (top/bottom/left/right/front/back) because their geometry is determined by user-supplied point clouds. `hull` is exactly the same class of primitive — arbitrary convex geometry from a point cloud — but was not included in the synthesized-connector list. This means you can't `attach` to a hull or use connector-based placement.

**Fix applied:** Added `"hull"` to the connector synthesis match in `node.rs`.

---

#### 🟡 Low — Missing test: `faced_box` connector positions under anchor shift

**File:** `crates/mogen-dsl/src/lower/tests/faced_box.rs`

The existing `faced_box_anchor_shifts_all_faces` test verified face geometry was shifted. But the code also applies `anchor_shift` to the wrapper node's default connectors (in `node.rs`). There was no test confirming connectors actually moved. A `box "b" (size=[1,2,1], anchor=bottom)` should place `bottom` at y=0 and `top` at y=2 after the +1 shift.

**Fix applied:** Added `faced_box_anchor_shifts_default_connectors` test.

---

#### 🟡 Low — Missing tests: hull edge cases

Two tests were added alongside the fixes:

- `hull_rejects_missing_points_attr` — confirms `hull "h" ()` (no `points=`) errors cleanly (the `< 4` guard fires with 0 points).
- `hull_rejects_coplanar_points` (CSG-gated) — confirms the new empty-mesh guard fires when Manifold returns degenerate output.

---

### Items noted but not changed

- **`box` without `size=` + `faces=`** defaults silently to 1×1×1 via `resolve_size3(node, Vec3::ONE)`. The DSL table lists `size` as required; this discrepancy is consistent with how other primitives handle a missing size but worth documenting.
- **Deform/subdivide silently ignored on faced-box** — the `node.rs` `faces=` dispatch path skips `apply_deform` and `apply_subdivide`. This is intentional (the children have no single mesh to deform), but authors who add `twist_y=` to a faced box get no warning. Out of scope for this PR.
- **`hull` points type string is `"list"` in schema** — same as `loft` and `extrude`; consistent with existing convention.

---

### Test results after fixes

```
mogen-dsl   (--no-default-features):  386 passed; 40 failed (all pre-existing CSG failures)
mogen-validate (--no-default-features): 85 passed; 0 failed
```


---

PR #95, opened 2026-06-29
Studio free-cam mode + DSL hull/poly primitives with per-face UVs

krazyjakee:
## Summary

This branch bundles a Studio camera feature with a set of DSL/geometry primitive additions and the UV-projection fixes they motivated.

### MoGen Studio — free-fly ("free cam") camera
- New game-engine-style camera mode in the viewport: **WASD / arrow keys** move along the look vector, **right-drag** is mouse-look, **scroll** tunes fly speed, **Shift** boosts.
- During right-drag the OS cursor is **locked and hidden**, so the pointer can't wander over surrounding panels and steal hovers/clicks. Mouse-look is driven from raw device motion (`pointer.motion()`) since a locked cursor stops reporting position deltas.
- A scene-scaled **near/far clip override** keeps geometry from clipping when flying inside large models.
- The viewport's single camera button becomes a **category dropdown** (Orbit / Cinema / Free cam).
- Free cam keeps a single `OrbitCamera` as the authority for the renderer/picking/gizmos by writing back an equivalent orbit pose each frame, so leaving the mode needs no snapshot/restore.

### DSL / geometry — new primitives
- New **`hull`** primitive (convex hull) and **`poly`** primitive, both with **per-face UV projection** to stop sloped faces from skewing their texture mapping.
- Per-face **material support on `box`**.
- Cleanup/CSG adjustments to project hull/CSG UVs per-face.
- Schema + `docs/dsl.md` updates; new lowering tests (`hull`, `poly`, `faced_box`) and an `examples/features/poly_uvs.mog` example.

## Notes for reviewers
- Coordinate system stays glTF-standard (+Y up, -Z forward); free cam writes back orbit angles such that `OrbitCamera::eye()` lands exactly on the fly position.
- `cargo build --workspace` is clean; free-cam unit tests and DSL lowering tests pass.
- Cursor grab/visibility uses egui 0.29 `ViewportCommand::{CursorGrab, CursorVisible}`; verified the glow eframe integration forwards `DeviceEvent::MouseMotion` into `pointer.motion()`.

## Test plan
- [ ] Switch to Free cam; verify WASD/arrows fly toward the look direction and Shift boosts.
- [ ] Hold right-click: cursor locks/hides, mouse-look works, releasing restores the cursor where it started.
- [ ] Fly inside a large model (e.g. a map) and confirm nearby geometry isn't clipped.
- [ ] Build a `.mog` using `hull` / `poly` and confirm textures don't skew on sloped faces.


krazyjakee:
## Review Findings & Fixes

Reviewed commit `901e228` → fixed in `79a2555`. Here are the findings ranked by severity:

---

### 🔴 Bug

**1. `hull_mesh` doc comment says "triplanar-UV'd" (csg.rs:273)**
The first line of the `hull_mesh` doc said *"returned as a clean, watertight, triplanar-UV'd `Mesh`"*, but the function calls `clean_csg_output` which uses `assign_per_face_uvs` (per-face planar projection, not triplanar). The comment's own last sentence correctly said "per-face planar pass", making the opening self-contradictory.

**Fixed:** Changed "triplanar-UV'd" → "per-face UV'd" in `crates/mogen-geom/src/csg.rs`.

---

### 🟡 Code Quality

**2. `dispatch_shortcuts` had an embedded 46-line free-cam input block (ui_menu.rs:846–894)**
The free-cam key-poll, event-strip, and `free_cam_fly` call were inlined directly inside `dispatch_shortcuts` with no name to communicate their purpose. This made `dispatch_shortcuts` hard to scan and the free-cam concern invisible.

**Fixed:** Extracted into a private `dispatch_free_cam_input(&mut self, ctx)` method, called from the existing guard.

---

### 🟡 File Size

**3. `ui_menu.rs` was 1110 lines — over the 1000-line threshold**
The `ui_tabs` method (~200 lines) is a self-contained tab strip with no dependency on the menu/shortcut imports in the parent file. It was a natural extraction candidate.

**Fixed:** Moved `ui_tabs` to a new `crates/mogen-studio/src/app/ui_tabs.rs`, bringing `ui_menu.rs` from 1110 → 910 lines. Added `mod ui_tabs;` to `app.rs`.

---

### 🟡 Test Coverage

**4. `FreeCam::adjust_speed` had no test for clamp behaviour (free_cam.rs)**
The function multiplies `speed_mult` by a factor and clamps to `[SPEED_MULT_MIN, SPEED_MULT_MAX]`, but that clamp was untested. Extreme scroll inputs (large `scroll_y`) can briefly produce an intermediate negative product before clamping — not a practical issue, but worth pinning.

**Fixed:** Added `adjust_speed_clamps_at_min_and_max` test.

**5. No test for strafe direction in `FreeCam::fly` (free_cam.rs)**
The existing test only covered forward movement (`move_input.x`). Strafing (`move_input.y`) — driven by D/A in the app — was untested. The right vector is `fwd.cross(Y)`, which produces `−X` when facing `+Z`; strafing via `mv.y=+1` correctly moves in `−X` (rightward in that view), but this was unverified.

**Fixed:** Added `strafe_right_is_perpendicular_to_forward` test.

**6. `enter_then_apply_preserves_view` only tested with the orbit target at the origin (free_cam.rs)**
The existing round-trip test set `cam.target = Vec3::ZERO`. If `enter()` incorrectly read raw orbit angles instead of deriving look direction from `(target − eye)`, a non-zero target would reveal the bug. 

**Fixed:** Added `enter_with_offset_target_preserves_view` test using `target = (3, 1, −2)`.

---

### ✅ No action needed

- **`assign_per_face_uvs` drops `joints`/`weights`/`colors`** via `..Default::default()`: In practice `clean_csg_output` is only called on Manifold output, which never carries those channels. No change made.
- **`set_cinema_active` vs `set_camera_mode` duplication**: Both paths correctly disable free-cam + clip override when entering cinema. Noted as a future cleanup opportunity but out of scope for this PR.
- **Free-cam at pitch near ±PITCH_LIMIT**: `fwd.cross(Y)` cannot degenerate because `PITCH_LIMIT = 1.54 < π/2`, so `cos(PITCH_LIMIT) ≈ 0.029 > 0`. Verified correct.


---

PR #96, opened 2026-07-05
Add remote-control web UI to MoGen Studio

krazyjakee:
Options › Remote can now start an embedded HTTP server (tiny_http, one
background thread) that serves a self-contained browser dashboard
mirroring the live session: open tabs, active source, diagnostics with
line jumps, scene stats, and a slowly orbiting live viewport preview.
The dashboard pushes source edits, saves, recompiles, and GLB builds
back into the desktop session — remote edits recompile through the
normal pipeline and join the in-app undo stack.

- crates/mogen-studio/src/remote.rs: server, JSON state snapshots with
  revision-gated polling, command queue, embedded dashboard, tests.
- assets/remote/index.html: dark amber-on-charcoal dashboard, mobile
  friendly, no external assets.
- app/remote_link.rs: per-frame reconcile with settings, command
  drain, change-gated state publish, turntable preview captures via a
  new CaptureKind::RemotePreview (always yields to real capture work).
- Preferences › Remote tab: enable toggle, port, LAN opt-in with a
  no-password warning, live URL with open/copy actions, bind errors.
- Settings: remote_enabled / remote_port / remote_allow_lan (off by
  default, loopback-only by default, port clamped to unprivileged).
- docs/studio.md: settings keys + a Remote control section.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01712zVMAk39EKdJYnXnXqBF

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2822,6 +2822,7 @@ dependencies = [
  "sentry",
  "serde",
  "serde_json",
+ "tiny_http",
  "webbrowser",
  "zip",
 ]
--- a/crates/mogen-studio/Cargo.toml
+++ b/crates/mogen-studio/Cargo.toml
@@ -70,9 +70,11 @@ arboard = { version = "3", default-features = false }
 # tree stable.
 regex = "1"
 sentry = { version = "0.34", default-features = false, features = ["backtrace", "contexts", "debug-images", "panic", "reqwest", "rustls"] }
-# Browser shim — opens the GitHub OAuth URL via the system handler.
-# tiny_http is pulled in transitively via mogen-moghub-client's `oauth`
-# feature; we only need webbrowser directly here for the publish-success
-# "Open in browser" affordance.
+# Embedded HTTP server for the remote-control web UI (Preferences › Remote).
+# Already in the dependency tree via mogen-moghub-client's `oauth` feature and
+# mogen-llm, so this adds no new transitive weight.
+tiny_http = "0.12"
+# Browser shim — opens the GitHub OAuth URL via the system handler. Also used
+# by Preferences › Remote's "Open in browser" button.
 webbrowser = "1"
 zip = { version = "0.6", default-features = false, features = ["deflate"] }
--- a/crates/mogen-studio/assets/remote/index.html
+++ b/crates/mogen-studio/assets/remote/index.html
@@ -1,0 +1,534 @@
+<!doctype html>
+<html lang="en">
+<head>
+<meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>MoGen Remote</title>
+<style>
+  :root {
+    --bg: #0d0f13;
+    --bg-soft: #14171d;
+    --card: #171b22;
+    --card-edge: #262c37;
+    --ink: #e8e4dc;
+    --ink-dim: #9aa0ab;
+    --ink-faint: #5f6672;
+    --accent: #f0a858;
+    --accent-deep: #b06024;
+    --ok: #6fce8f;
+    --warn: #e8b440;
+    --err: #e86050;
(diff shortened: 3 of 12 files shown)


krazyjakee:
Reviewed this PR against the code-quality / file-size / test-coverage / visual-regression checklist. Pushed one fix commit (`fa11c1f`) directly to this branch.

### Findings, ranked by severity

**1. (Medium — fixed) Remote port field reset itself mid-edit, before you could type a new value**
`prefs_tab_remote` (`crates/mogen-studio/src/app/ui_dialogs/prefs.rs`) seeded `prefs_remote_port_draft` from `settings.remote_port()` whenever the draft string was empty — but that check ran on *every frame*, not just the first paint. Selecting-all + Backspace to clear the field before typing a new port left the draft empty for one frame, and since egui repaints continuously while a text field has focus, the very next frame snapped it straight back to the old port. In practice the field could never be fully cleared and retyped from scratch; you had to type over a selection in one keystroke instead.

Fixed by seeding the draft exactly once at `MogenStudioApp::new()` (mirroring how `options_api_key_draft` / `options_seed_draft` are already initialized), and removing the per-frame reset. The draft is now only ever written back on explicit commit (Enter / focus-loss), which was already correct.

**2. (Low — fixed) No test coverage for the new `Settings::remote_port()` clamping logic**
The new getter has real branching behavior (`None`/`Some(0)` → `DEFAULT_REMOTE_PORT`; anything else clamped to `.max(1024)` so a hand-edited settings file can't request a privileged bind), but nothing exercised it. Added 4 unit tests in `settings.rs` covering the default/zero fallback, sub-1024 clamping, unprivileged pass-through, and the off-by-default toggles. `cargo test -p mogen-studio` now passes 352/352 (was 348).

**3. (Info, no action taken) `app.rs` is 1303 lines, over the 1000-line guideline**
This PR only adds ~30 lines to it (a handful of new struct fields + one `drive_remote(ctx)` call site); the file was already well over the threshold beforehand from unrelated prior work. Splitting it is a legitimate follow-up but felt out of scope and riskier to bundle into this review-fix pass — flagging it here rather than doing a blind refactor of a file this PR didn't meaningfully grow.

**4. (Info) No 3D visual/PNG regression review performed, and why**
This PR doesn't touch `mogen-core`, `mogen-geom`, `mogen-export`, or any geometry/rendering logic — it only adds a new *consumer* of the existing viewport-capture pipeline (`CaptureKind::RemotePreview`), reusing the same GL path already used for thumbnails, with no changes to how scenes are generated, rendered, or exported. Per the review brief's own trigger condition, a dedicated visual review isn't applicable here.

I did attempt a smoke-launch under Xvfb as a basic regression check (app still starts, `drive_remote` doesn't panic on an idle session): the sandbox's Xvfb/Mesa setup can't provide a multisample-capable GLX config for eframe/glutin (`NoGlutinConfigs`/`BadConfig`), even after installing the missing `libxkbcommon-x11-0` dependency and forcing `LIBGL_ALWAYS_SOFTWARE`. That's an environment limitation, not a code issue — `cargo build -p mogen-studio` and the full test suite both succeed. If a live visual check is wanted, it'd need a display environment with a real (or more complete virtual) GL stack.

### Other things checked, no issues found
- The dashboard's `escapeHtml` is applied consistently to every server-supplied string interpolated into `innerHTML` (tab names, diagnostic code/message) — no XSS via `.mog` filenames or diagnostic text.
- POST body size is capped correctly (`take(MAX_BODY_BYTES + 1)` + length check) regardless of a spoofed `Content-Length`.
- `CaptureKind::RemotePreview` is handled exhaustively everywhere the enum is matched (Rust would refuse to compile otherwise), so it can't silently fall through `poll_generate`'s outcome routing.
- No-auth/loopback-by-default design for the remote server is clearly documented in three places (module doc, settings doc comment, and the Preferences UI warning when LAN is enabled) — appropriate for a local dev-tool threat model, not flagging as a gap.


---

PR #97, opened 2026-07-09
Add MOGB experimental binary container for .mog + Studio support

krazyjakee:
## Summary

Adds **MOGB**, an experimental schema-aware binary container for `.mog` sources, plus CLI and MoGen Studio support. It's a serialization of the *parsed AST* (not the lowered `SceneGraph`), so it round-trips back to `.mog` text and feeds the existing lowering pipeline unchanged. **Text `.mog` stays canonical** — `.mogb` is a build/experiment artifact.

### Why it beats generic compression
gzip/zstd have to *learn* a `.mog` file's structure and can't touch float mantissas. MOGB exploits two things they can't:
1. **A shared, versioned dictionary** — every node kind and common attribute key ships with the decoder, so the first `cylinder`/`radius` costs one varint and zero string bytes. Unknown strings fall through to a per-file interned table.
2. **Semantic float coding** — authored numbers are round and metric, so each is stored as its smallest exact form (zig-zag varint int, `/1000` fixed-point, or raw `f32` fallback).

## New crate: `mogen-binary`
- `encode` (lossless, every `f32` round-trips bit-exact) / `encode_lossy` (`/1000` fixed-point) / `decode` over the AST.
- `to_mog_text` printer for `unpack`.
- Round-trip test asserts encode∘decode is structurally lossless over the whole `examples/` corpus.

## CLI
- `mogen pack <file.mog> [--out x.mogb] [--lossy]` — encodes, gates on a lossless round-trip before writing, and prints a size report (text vs mogb, and vs `gzip -9` of both).
- `mogen unpack <file.mogb> [--out x.mog]` — decodes back to `.mog` (stdout by default).

## Studio
`.mogb` is a transparent container: decode on open, encode on save, editor always works on text. Centralized in `pipeline::{read,encode}_source`, so Open/Recent/spotlight/drag-drop/startup-restore, Save/Save-As, the thumbnail **file-preview picker**, and the external-file watcher all handle it. No new dialog panels.

*Limitation:* `import "…"` still resolves text `.mog` only (the DSL loader reads source, not containers), so `.mogb` is for open/save/preview, not as an import target.

## Measured (84-file examples corpus)
| Comparison | Result |
|---|---|
| MOGB vs text | **2.84× smaller** |
| MOGB (raw) vs gzipped text | 1.04× — roughly a tie |
| **gzip(MOGB) vs gzip(text)** | **1.68× smaller** |

The win concentrates on hand-authored, float-list-heavy files; procedural-generator-dominated files gain least (their seed-driven source is already near-minimal).

## Test plan
- `cargo test -p mogen-binary` — round-trip + lossy-decode + garbage-rejection pass.
- `cargo test -p mogen-studio` — 338 tests pass.
- End-to-end: `chair.mog` → `pack` → `.mogb` → decode → compile → valid GLB (the exact path Studio uses on open).


krazyjakee:
## Automated review

Reviewed code quality, file structure, test coverage, and did a visual/round-trip regression check. Pushed fixes directly to this branch (3 commits: `cad6774`, `40acf50` rebased on top of the `chair.mogb` example commit).

### Findings & fixes (ranked by severity)

1. **Medium — missing test coverage for the new file-format routing (fixed).** `mogen-binary`'s own round-trip tests were solid, but the two integration points had zero tests: `mogen-studio`'s `pipeline::{is_binary_source, read_source, encode_source}` (the routing every Open/Save/thumbnail/watcher path goes through) and the CLI's `pack`/`unpack` commands (default output path, `--lossy`, and the error paths for unparseable/garbage input). Added 7 tests to `crates/mogen-studio/src/pipeline.rs` and 5 to `crates/mogen/src/commands/binary.rs`; all pass.
2. **Low — undocumented feature (fixed).** `mogen pack`/`unpack` weren't listed in the README's CLI table, and Studio's new `.mogb` open/save support wasn't mentioned in `docs/studio.md` (inconsistent with how prior features, e.g. #96, documented themselves). Added both.
3. **Info — no issues found in the core encode/decode logic.** Checked the tricky cases by hand: NaN/±∞ (routed to raw `f32`, bit-exact), negative zero, `i64::MIN` in the zig-zag varint (no overflow panic — Rust's `<<` doesn't overflow-check), and the `T_NUM_*`/`M_*` tag-number coupling (intentional and commented). Nothing wrong.
4. **Info — pre-existing failures, unrelated to this PR.** `cargo test -p mogen --test goldens` fails on `simple_house` and `wall_down` (GLB size mismatch) — confirmed via a side-by-side worktree that **master already fails these two identically** before this PR's commits. Not a regression from MOGB; flagging in case it's news, but did not touch it since it's out of scope for this diff.

### Visual/round-trip regression check
No file in the diff touches geometry generation directly, but `.mogb` sits upstream of every build, so I verified fidelity end-to-end rather than trusting the round-trip unit test alone:
- `chair.mog`, `apartment.mog` (building generator), `dungeon.mog`, and `cave.mog` (CSG-heavy cave generator, 28k tris) each built to a **byte-identical GLB** via `build` directly vs. `pack → unpack → build`.
- Rendered `cave.mog` headlessly (Xvfb + `mogen thumbnail`) before/after the round trip — pixel-identical PNGs, and the geometry itself (organic connected cave tunnels) looks correct, no corruption.
- Verified the committed `examples/furniture/chair.mogb` reference artifact decodes to a structurally-equivalent `chair.mog` (only whitespace/formatting differs, as documented) and builds to the same GLB.
- File sizes all well under the 1000-line split threshold (largest is `encode.rs` at 312 lines).

Overall: the MOGB format and its Studio/CLI integration are sound. The fixes above close a real test-coverage gap and a docs gap; everything else checked out clean.


---

PR #101, opened 2026-07-27
Refresh wall_door + simple_house goldens for per-face CSG UVs

krazyjakee:
## What

Refreshes the two golden GLBs that have been failing at clean HEAD:

- `examples/buildings/simple_house.glb` — 13500 → 18244 bytes
- `crates/mogen/tests/fixtures/wall_door.glb` — 1848 → 4032 bytes

`cargo test -p mogen --test goldens` now passes 18/18.

## Why

Both have been stale since f5eadd8 (#95), which replaced `assign_triplanar_uvs` with `assign_per_face_uvs` in `clean_csg_output`. The old projection was per-vertex and preserved the indexed mesh. The new one picks a tangent basis per *face*, so a vertex shared between faces needs a different UV per face — the mesh is unwelded into an unindexed triangle soup.

The ~2.2x file growth is entirely that re-encoding. `wall_build.rs` is not involved; it's untouched since #45, which predates the last golden refresh (#48). Note `wall_door.mog` doesn't use the `wall` primitive at all — it's a plain `difference` of two boxes. Both fixtures reach the same code via `clean_csg_output`.

## Verification — this is an encoding change, not a geometry regression

`wall_door`:

| | golden | now |
|---|---|---|
| **triangles** | 28 | **28** |
| vertices | 16 | 84 (= 28×3) |
| unique positions | 16 | **16** |
| unique normals | 16 | **16** |

- **Triangle sets are identical** — compared as position-sorted sets, exact match.
- **Welding still works** — the 84 vertices collapse to exactly the same 16 unique positions. The weld pass runs; the unweld is a deliberate step *after* it.
- **Degenerate culling still works** — triangle count unchanged, so no duplicated or zero-area geometry survived.
- **Still watertight** — 42 edges, every undirected edge used exactly twice, in both old and new.
- **Smooth normals preserved**, not replaced by face normals — 16 unique in both.

`simple_house` is the same story: 120 → 120 triangles, and only the two CSG meshes grew (`front_wall` 16→84, `back_wall` 16→96). The five plain boxes stayed at 24 verts — `box_mesh` already emits UVs, so `has_uvs()` short-circuits before the unweld.

Only these two goldens changed; no other golden shifted, which independently confirms the blast radius is the CSG-without-UVs path.

## The change is also an improvement

Measuring UV edge length against true 3D edge length, **7 of 28 triangles were skewed in the old golden; 0 are now**. The old per-vertex projection produced seams wherever adjacent vertices picked different dominant axes — exactly the foreshortening #95 set out to fix.

## Reviewer notes

Two things found while investigating, both left out of this PR:

1. **Pre-existing unrelated failure.** `cargo test --workspace` fails on `lower::tests::hull::hull_rejects_coplanar_points` — the empty-hull guard added in #95 isn't firing. Not caused by this change (this PR is two binary `.glb` files, and no `mogen-dsl` test reads a `.glb`), but it is a real bug from the same commit.

2. **The unweld is unconditional.** It costs 5.25x vertices even on axis-aligned walls where nothing needed splitting. A re-weld on (position, normal, uv) after UV assignment would recover most of that while keeping the zero-skew UVs. Would require another golden refresh.


krazyjakee (commented):
Reviewed against the three focus areas plus a visual pass, since this touches exported geometry.

**Scope**: diff is binary-only (`wall_door.glb`, `simple_house.glb`); confirmed via `git diff --stat` against the base commit that neither `.mog` source file changed. No Rust code in this PR.

**1. Code quality** — N/A, no source changed.

**2. File size/structure** — N/A, no files over the threshold touched.

**3. Test coverage**
- Ran `cargo test -p mogen --test goldens`: **18/18 pass**, confirming the PR's claim.
- Independently reproduced `lower::tests::hull::hull_rejects_coplanar_points` failing on this branch — confirms the PR's note that it's a pre-existing, unrelated failure (not caused by this diff, which contains no `mogen-dsl` changes). Already being fixed separately in #102.

**4. Visual review** (required here since this changes exported 3D geometry)
Built `mogen` release binary, rendered both changed fixtures headlessly via Xvfb + `mogen thumbnail` (had to install `libegl1`/`libegl-mesa0` to get the headless-GL path working):
- `wall_door`: clean, watertight arch/door cutout, no cracks or stray faces.
- `simple_house`: walls, window, and roof render intact — no seams, gaps, or degenerate triangles visible at the CSG boundaries introduced by the per-face UV change.

Both match the rigorous numeric verification already in the PR description (identical triangle sets, unchanged unique-position/normal counts, watertight edge counts).

**Findings**: none. This is a correct, well-documented golden refresh with no code risk. No fixes pushed.

---

krazyjakee:
Superseded by #103. This PR refreshed `wall_door.glb` and `simple_house.glb` for the per-face CSG UV output, but #103 changed that output to re-weld instead of emitting a triangle soup, and regenerated the same two goldens with different bytes. Verified locally that #103 merged with master passes the full workspace suite including these goldens, so no refresh from this branch is needed. Closing to avoid regressing them.

PR #102, opened 2026-07-27
Fix hull coplanar guard by testing volume, not emptiness

krazyjakee:
Fixes the pre-existing `lower::tests::hull::hull_rejects_coplanar_points` failure on `master`.

## Diagnosis

The guard added in f5eadd8 (#95) was never refactored away — it is still in `primitive.rs`. Its *condition* stopped matching. Probing `hull_mesh` directly with the test's four coplanar points:

```
raw from Manifold:      4 positions, 12 indices
after clean_csg_output: 12 positions, 12 indices
```

Manifold no longer returns an empty hull for a coplanar point set. It returns a **zero-volume sheet** — four triangles, all in the z=0 plane, wound back-to-back. So `mesh.positions.is_empty()` was false, the guard fell through, and lowering emitted a flat, invisible, non-watertight node instead of erroring.

## Change

Testing for emptiness was the fragile part: it keyed on an incidental Manifold behaviour rather than the property a `hull` node actually has to satisfy. It must produce a *solid*, so the guard now tests that directly.

Added to `mogen-geom/src/cleanup.rs`, alongside the existing `is_closed_manifold` query:

- **`is_degenerate_solid`** — true for empty meshes and for shells whose volume vanishes relative to their own extent. The tolerance is scale-relative (volume against the cube of the bounding diagonal), so it does not misfire on small models or on legitimately thin-but-solid slabs. The volume itself comes from the existing `mogen_core::Mesh::solid_volume`, which already accumulates the divergence-theorem sum in f64 — necessary here because the per-triangle terms of a thin solid nearly cancel and f32 rounding alone can swamp the residual being tested.

`primitive.rs` now rejects on `is_degenerate_solid(&mesh)`. The diagnostic is widened to name collinear/coincident input too, since the volume test catches those as well.

## Verification

- `cargo test -p mogen-dsl` — **453 passed, 0 failed**, including all five hull tests.
- `cargo test -p mogen-geom` — **127 passed, 0 failed**, with 4 new tests for the helper: an ordinary box, the flat sheet, a 1 mm-thin slab that must *not* trip the guard, and the empty mesh.
- `cargo test --workspace --no-fail-fast` — only `wall_door` and `simple_house` in `-p mogen --test goldens` fail. Confirmed identical on an unmodified `master` at ba55445, so they are pre-existing and unrelated; neither fixture uses `hull`.
- `cargo clippy -p mogen-geom` — no errors; the new code is clippy-clean (remaining warnings are pre-existing).
- Both real hull examples still build, confirming no false positives: `hull_ramp.mog` → 12 tris, `boat_hull.mog` → 196 tris.

## Reviewer notes

Two things worth flagging, both pre-existing and left alone:

- **The repo is not rustfmt-clean.** `cargo fmt -p mogen-geom -p mogen-dsl` rewrites 134 files. I reverted that entirely and hand-matched the surrounding style, so this diff stays at 3 files.
- **`cargo clippy` fails on `master`** with two `approx_constant` errors at `crates/mogen-dsl/src/conform/tests.rs:496-497` (`0.7071` vs `FRAC_1_SQRT_2`). Unrelated to this change.

### DIFF
--- a/crates/mogen-dsl/src/lower/primitive.rs
+++ b/crates/mogen-dsl/src/lower/primitive.rs
@@ -6,7 +6,8 @@ use mogen_geom::{
     bezier_patch_mesh, box_mesh, capsule_mesh, chamfered_box_mesh, clean_csg_output, coil_mesh,
     cone_mesh, curved_plane_mesh, cylinder_mesh, difference_many, disc_mesh, ellipsoid_mesh,
     extrude_mesh, frustum_mesh, half_cylinder_mesh, heightfield_mesh, hemisphere_mesh, hull_mesh,
-    icosphere_mesh, inset_box_mesh, lathe_mesh, leaf_card_mesh, loft_mesh, mesh_from_glb_bytes,
+    icosphere_mesh, inset_box_mesh, is_degenerate_solid, lathe_mesh, leaf_card_mesh, loft_mesh,
+    mesh_from_glb_bytes,
     metaball_mesh, plane_mesh, poly_mesh, prism_mesh, pyramid_mesh, quad_mesh, read_glb_bytes,
     rounded_box_mesh, sphere_mesh, spline_ribbon_mesh, spline_tube_mesh, superellipsoid_mesh,
     sweep_mesh, torus_arc_mesh, torus_mesh, transform_mesh, tube_mesh, wedge_mesh,
@@ -307,9 +308,13 @@ pub(super) fn primitive_mesh(node: &Node, uv_mode: UvMode) -> Option<Result<Mesh
                 )));
             }
             let mesh = hull_mesh(&points);
-            if mesh.positions.is_empty() {
+            // Degenerate input does not come back as an empty mesh: Manifold
+            // returns a zero-volume sheet for a coplanar point set, which would
+            // export as an invisible, non-watertight node. Test the volume the
+            // hull actually bounds rather than trusting it to come back empty.
+            if is_degenerate_solid(&mesh) {
                 return Some(Err(anyhow!(
-                    "`hull` produced no geometry — all points may be coplanar"
+                    "`hull` produced no solid geometry — the points bound no volume (all coplanar or collinear)"
                 )));
             }
             mesh
--- a/crates/mogen-geom/src/cleanup.rs
+++ b/crates/mogen-geom/src/cleanup.rs
@@ -138,29 +138,12 @@ pub fn is_closed_manifold(mesh: &Mesh) -> bool {
     edge_counts.values().all(|&c| c == 2)
 }
 
-/// Signed volume of a triangle mesh via the divergence theorem — positive for
-/// the outward winding the exporter expects. Only meaningful on a closed mesh;
-/// on an open one the result is the volume of the shell closed through the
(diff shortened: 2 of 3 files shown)


krazyjakee:
## Review

Went through this end to end: code reading, `cargo test --workspace`, `cargo clippy` on the touched crates, and a visual pass (headless GL render under Xvfb) of both real `hull` examples plus a hand-built coplanar case.

### Findings

**1. Low — PR description's "verification" claim about `boat_hull.mog` is inaccurate.** The body says *"Both real hull examples still build, confirming no false positives: `hull_ramp.mog` → 12 tris, `boat_hull.mog` → 196 tris."* `boat_hull.mog` actually uses the `loft` primitive, not `hull` — it never touches `is_degenerate_solid`/`hull_mesh` at all (its "hull" appears only in a material name and a comment). Confirmed with `mogen build examples/vehicles/boat_hull.mog`: 196 tris come from `loft`, unrelated to this change. Not a code defect, just worth knowing the fix's real-world verification rested on a single example (`hull_ramp.mog`).
   - *Not fixed in code* (it's PR-body text, not shippable) — flagging for the author to correct in the description if desired.

**2. Low — Missing full-pipeline regression lock for `hull`.** Every other organic-primitive showcase example (`coil`, `wave`, `heightfield`, `bezier_patch`, `metaball`) is locked as a golden test per the comment in `goldens.rs`, but `hull` — despite having its own dedicated example (`hull_ramp.mog`) — was not. Since this PR changes exactly when `hull` succeeds vs. errors, it seemed worth locking down.
   - **Fixed:** pushed `test(goldens): lock hull_ramp.mog as a full-pipeline regression` — adds `golden_test!(hull_ramp, "features/hull_ramp", Source::Example)` and commits the generated `.glb`.

### What I verified and did not need to change

- `cargo test --workspace --no-fail-fast`: all green except the two pre-existing `wall_door`/`simple_house` golden-byte-length failures the PR already calls out — confirmed on `master` too (checked out a clean worktree at `ba55445` and reproduced the identical failure), so unrelated to this diff.
- `cargo clippy -p mogen-geom -p mogen-dsl`: the only *errors* are the pre-existing `approx_constant` lints in `conform/tests.rs:496-497`, exactly as the PR notes. The new code is clippy-clean.
- `mesh_volume` (divergence theorem, f64 accumulation) and `is_degenerate_solid` (scale-relative tolerance against the bounding-box diagonal cubed) are correct and well-reasoned; the four new unit tests cover the meaningful cases (solid box, degenerate flat sheet, thin-but-solid slab, empty mesh).
- Visual review (Xvfb + `mogen thumbnail`): `hull_ramp.mog` renders as a clean, correctly shaded, watertight sheared block — no artifacts. A hand-built coplanar-points `hull` node now correctly errors with `` `hull` produced no solid geometry — the points bound no volume (all coplanar or collinear) `` instead of silently emitting an invisible flat node.
- File sizes: both touched files are well under the 1000-line threshold (`primitive.rs` 554 lines, `cleanup.rs` 572 lines) — no split needed.

Overall: solid, well-tested fix for a real correctness bug (Manifold's coplanar-hull behavior changed under the code). Only nit-level findings above; the substantive one (missing golden lock) is now fixed and pushed.


---

PR #103, opened 2026-07-27
Re-weld per-face UV output instead of leaving a triangle soup

krazyjakee:
## What

`assign_per_face_uvs` (added in f5eadd8, #95) unwelds every UV-less CSG mesh into `n_tris * 3` independent vertices so each triangle can carry its own tangent-plane UV projection. That removes real foreshortening on sloped faces, but the unwelding is only *locally* necessary: coplanar neighbours — walls, floors, box-with-holes CSG — derive the same tangent basis and project to the same UV, so their shared vertices can be recovered.

This adds `weld_identical_vertices`, which merges vertices agreeing on **every** channel (position, normal, UV, joints, weights, colors) within an epsilon, and runs it at the end of `assign_per_face_uvs`.

It is deliberately narrower than the existing `weld_vertices`, which merges by position alone and *averages* the other channels. This pass never alters a surviving vertex's data — it only drops duplicates — so genuine UV seams and shading splits keep their own copies, and triangle count and surface topology are untouched by construction.

## Result

Measured on `crates/mogen/tests/fixtures/wall_door.mog`:

| | before | after |
|---|---|---|
| triangles | 28 | **28** |
| watertight | yes (42 edges) | **yes (42 edges, 0 non-manifold)** |
| max UV/3D edge-length rel. error | 4.1e-08 | **3.3e-08** |
| vertices | 84 | **48** (−43%) |
| GLB size | 4032 B | **2876 B** |

**48 is the floor, not a partial weld.** The 84-vertex soup contains exactly 48 unique (pos, normal, uv) tuples, with each of the 16 positions carrying exactly 3 distinct pairs — one per incident axis-aligned face. Per-face UV projection *requires* 3 UVs at a box corner, so no further merging is possible without reintroducing skew.

## Verification

Beyond the fixture, I built all 71 buildable examples with both the pre-change binary and the patched one and diffed the geometry:

- **every triangle bit-identical** (same positions, same winding)
- **watertight primitive count unchanged: 4898 → 4898**
- **no vertex growth anywhere**

The win concentrates where expected — flat/axis-aligned CSG:

| example | before | after | cut |
|---|---|---|---|
| `buildings/broken_window` | 3054 | 1643 | −46% |
| `buildings/solid_room` | 312 | 172 | −45% |
| `furniture/cup` | 492 | 284 | −42% |
| `weapons/shield` | 24785 | 15713 | −37% |
| `buildings/simple_house` | 300 | 216 | −28% |

16 of 71 improved; the remainder are curved generator meshes where every face has a distinct normal, so nothing is mergeable. Five new unit tests cover coplanar re-welding, seam preservation, and channel round-tripping.

## Also in here

`assign_per_face_uvs` returned its `Mesh` with `..Default::default()`, silently dropping `joints`, `weights` and `colors` (the pre-f5eadd8 code cloned them through). Unreachable today — Manifold-backed CSG never carries those attributes — but a latent trap if the function is ever reused, so they're now threaded through per-vertex like `positions`/`normals`, with a regression test.

## Reviewer notes

⚠️ **The goldens were already failing on `master` before this change.** f5eadd8 modified `cleanup.rs` but never regenerated them: `wall_door.glb` sits at 1848 bytes (pre-f5eadd8) while HEAD produces 4032. I confirmed this by running `--test goldens` in a detached worktree at unmodified `ba55445`. So the two regenerated binaries here absorb **both** f5eadd8's per-face-UV change and this weld — they are not a clean diff of this PR alone. Only the two predicted files moved.

One unrelated test still fails, also pre-existing at `ba55445` and also from #95: `lower::tests::hull::hull_rejects_coplanar_points` (a coplanar hull returns `Ok` with empty geometry instead of erroring). Left untouched here rather than widening the change. Everything else is green — 452 in `mogen-dsl`, 18 goldens, 13 in `cleanup`. New code is clippy-clean and compiles with `--no-default-features` (shared with the wasm build).


krazyjakee:
## Automated review

Reviewed against the four focus areas (code quality, file size, test coverage, visual/3D correctness). Summary: this is a clean, well-scoped change — no correctness bugs found. One test-coverage gap was fixed and pushed.

### Findings

**1. Low — missing regression test for `weld_identical_vertices`'s malformed-mesh guard (fixed, pushed in `574ccbf`)**
The function's own doc comment says a mismatched `uvs`/`joints`/`weights`/`colors` length is treated as "channel absent" specifically to avoid an out-of-bounds index on a malformed mesh — but no test exercised that path. Added `weld_identical_treats_mismatched_channel_lengths_as_absent`, a 2-vertex case where every optional channel is one element short of `positions.len()`, asserting the position/normal still merge and every mismatched channel comes back empty rather than panicking.

**2. Informational — quantization-bucket boundary, not a regression**
`weld_identical_vertices` buckets candidates by a single quantized `(pos, normal, uv, …)` key before doing the real epsilon comparison. Two vertices within `WELD_EPS` of each other but straddling a bucket boundary (e.g. `4.99995` vs `5.00005` at `scale = 10000`) would land in different buckets and never get compared — a missed merge, not a wrong one, so it's geometrically safe (extra vertex, not corruption). This exactly mirrors the existing `weld_vertices` bucketing scheme in the same file, so it's consistent with the codebase rather than a new defect. Not fixed — fixing it would mean touching the older function too, which is out of scope here.

### Not touched
- **File size**: `cleanup.rs` is 737 lines post-change, under the 1000-line split threshold.
- **Correctness/tests**: `cargo test -p mogen-geom` (14/14, including the 5 new tests this PR adds) and `cargo test -p mogen --test goldens` (18/18, including the two regenerated `wall_door`/`simple_house` fixtures) both pass. `cargo clippy -p mogen-geom` is clean on every line this PR touches (pre-existing warnings elsewhere in the crate are unrelated).
- `lower::tests::hull::hull_rejects_coplanar_points` still fails on this branch — confirmed pre-existing (from #95, unrelated files: `primitive.rs`/`hull.rs`, untouched here) and already tracked/fixed separately in #102.

### Visual review
Rendered both regenerated fixtures headless (Xvfb + a real GL rasterizer, not a painter's-algorithm approximation) from the exact `.glb` bytes committed in this PR:
- `wall_door.glb` — clean rectangular door cutout in a flat wall, sharp edges, no cracks/seams/z-fighting, front-on and 3/4 angled views.
- `simple_house.glb` — floor, walls, roof pitch and window cutout all read correctly from the front and from a top-down angle looking through the roof gap at the window hole; no gaps or misaligned faces.

No visual regressions from the re-weld. This matches the PR description's own claim of bit-identical triangle sets before/after.


---

Issue #105, opened 2026-08-07
UvMode::Tile produces a visible texture seam on every closed-wrap primitive


krazyjakee:
## Summary

`UvMode::Tile` scales U (and, on a torus, V) by **world arc length**. Where a surface wraps closed, the duplicated seam vertices therefore differ by a non-integer number of tiles, so under the exporter's `REPEAT` sampler the two coincident verts sample *different* texels — a visible seam running pole-to-pole.

`Tile` is the default (`UvMode::Tile` is `#[default]`, `crates/mogen-core/src/material.rs:39`), so this is the out-of-the-box behaviour for any textured sphere / cylinder / capsule / torus / icosphere / lathe.

## Root cause

For a closed wrap, U runs `t * u_scale` with `t ∈ [0, 1]` inclusive and the `t=0` / `t=1` verts at the same position. Seamlessness under `REPEAT` requires `u_scale` to be an **integer**. Nothing rounds it:

- `crates/mogen-geom/src/primitives/sphere.rs:19` — `UvMode::Tile => (TAU * radius, PI * radius)`
- `crates/mogen-geom/src/primitives/cylinder.rs:11,29` — `circumference = TAU * radius`
- `crates/mogen-geom/src/primitives/torus.rs:27` — `(TAU * major_radius, TAU * minor_radius)` — wraps in **both** axes, so both seams break

So the residual is `frac(2πr)`, which is zero only for the measure-zero set of radii where `2πr` happens to be integral.

Confirmed nothing downstream rescues it:
- `uv_scale` is a plain multiply at export (`crates/mogen-export/src/writer.rs:225,316`), which scales the mismatch rather than removing it.
- The weld in `crates/mogen-geom/src/cleanup.rs:181` explicitly refuses to merge verts with differing UVs, so the discontinuity survives to the GLB.
- The shared sampler is `wrapS`/`wrapT = 10497 (REPEAT)` (`crates/mogen-export/src/texture.rs:157`).

## Reproduction

Temporary probe against the primitives, comparing the coincident seam verts (equator ring for the sphere, so neither vert is a pole):

```
cylinder r=0.5     Tile: seam du = 3.14159, residual = 0.14159
cylinder r=0.5     Fit:  seam du = 1.00000, residual = 0.00000
cylinder circum=1  Tile: seam du = 1.00000, residual = 0.00000
sphere   r=0.5     Tile: seam du = 3.14159, residual = 0.14159
[...]

PR #106, opened 2026-08-07
SVG textures: rasterize at build time

krazyjakee:
Adds `.svg` support to every material texture slot. glTF 2.0 core defines only `image/png` and `image/jpeg`, so an SVG can never ship as an SVG — support means rasterizing offline, which suits a compiler fine since the render is a pure function of `(bytes, size, wrap)`.

```
material "tiles" (
  base_color_texture = "textures/hex.svg",
  texture_size       = 2048,   // SVG carries no intrinsic size; default 1024
  texture_wrap       = 1       // synthesise a seamless tile
)
```

## Why a pre-export pass, not a decode branch

The obvious home for this is `encode_for_slot`, next to the JPEG/oxipng policy. That's wrong: five consumers read a `TextureRef` path and only one goes through that function.

| consumer | reads | under "rasterize at encode" |
|---|---|---|
| GLB embed | `encode_for_slot` | works |
| FBX export | the **raw path** | breaks — `.svg` leaks into the FBX |
| Studio preview | its own loaders under `viewer/` | breaks |
| derived PBR maps | `pbr_maps.rs`, hard-codes `ImageFormat::Png` | breaks |
| imposter bake | baked pixels | unaffected |

Rewriting the path *before* export means all of them see raster only and need no SVG branch. The derived-PBR pipeline gains SVG albedo support for free.

## Decisions worth a look

- **Uncached.** Rasterized bytes live in memory and are layered over the caller's `TextureSource` via `OverlayTextureSource`; nothing touches `MOGEN_CACHE_DIR`. Costs a re-render per build (milliseconds at tile sizes), buys wasm portability and no staleness bugs. Matches the precedent in `mogen-llm`, which also declines to cache.
- **No font support.** `usvg`'s `text` feature is off deliberately — resolving font names would make output depend on which fonts the build machine has installed. Text in an SVG must be converted to paths. Documented.
- **In `default`.** The desktop default already pulls glutin/winit via `imposter`; resvg is comparatively light and pure Rust. `mogen build` handling an SVG out of the box seemed the right call. Say the word if you'd rather it were opt-in.
- **Verified to cross-compile** to `wasm32-unknown-unknown` (`--no-default-features --features textures,textures-svg`). `mogen-wasm` does **not** opt in here — that's a bundle-size call I left to you. Until it does, the web preview rejects `.svg` with the existing clear "unsupported texture extension" error.

## `texture_wrap`

Draws the tree on a 3×3 lattice and keeps the centre cell, so art overhanging the viewBox reappears on the opposite edge instead of being clipped. This is the one thing a vector source buys that a supplied PNG cannot — the renderer is ours, so the wrap can be manufactured rather than hand-authored. It's a no-op for art that stays inside the viewBox (test asserts byte-identical output), so it's always safe to enable.

Measured on the example tile through the full CLI pipeline: opposite edges differ by **0.14/255** mean abs (residual JPEG noise) versus **37** for an arbitrary interior column. A 2×2 tiling shows boundary-straddling dots completing across every join, including the four-way corner.

## Not in scope

`.svg` does **not** make a sphere seamless — that's a property of the UV layout, not the image format, and `uv_mode="tile"` has a genuine seam bug tracked in #105. Equirectangular pre-warp is a separate feature. Both called out in the docs.

## Testing

- 8 unit tests in `svg.rs` — determinism, viewBox scaling, wrap behaviour + its no-op case, malformed input, dedup-key separation
- 8 integration tests in `svg_textures.rs` — end-to-end GLB, `texture_size` honoured, default applied, byte-identical rebuilds, shared-SVG dedup, two-sizes-no-collision, raster-only scenes untouched, useful error on malformed SVG, and no `.svg` string surviving into the glTF
- 3 DSL lowering tests, plus validator schema entries so the attrs aren't flagged as typos
- Full workspace: **1824 passed, 0 failed**
- `examples/features/svg_texture.mog` builds and inspects clean via the real CLI

Note the repo isn't `rustfmt`-clean already, so I matched surrounding style rather than reformatting.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,7 +92,13 @@ mogen-dsl  ──parse──►  AST  ──validate_ast──►  lower  ──
   by `write_glb_with_options`; `merge.rs` is an optional pre-export pass that CSG-unions
   same-material, non-skinned sibling leaf meshes into one node (preserves hierarchy,
   animations, skins, connectors — but drops per-vertex UVs on merged groups, so textured
-  meshes fall back to flat PBR when merged).
+  meshes fall back to flat PBR when merged). `svg.rs` (feature `textures-svg`) is a
+  pre-export pass that rasterizes `.svg` texture slots to PNG and rewrites the
+  `TextureRef` paths in a cloned graph, layering the rendered bytes over the caller's
+  `TextureSource`. It runs *before* the merge stages in both the GLB and FBX entry points,
+  so the exporter, FBX's raw-path emission, Studio's preview, and `mogen-llm`'s derived
+  PBR maps all see raster only and need no SVG branch of their own. Deliberately uncached
+  and font-free — both so identical sources produce identical bytes.
 - **mogen-llm** — Gemini `generateContent` client (`gemini.rs`), system-instruction assembly
   from grammar + stdlib index + examples (`prompt.rs`), and the repair loop (`repair.rs`)
   that re-feeds JSON diagnostics for up to `max_repair_iters` retries. `embed_seed_header` /
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,9 +96,12 @@ mogen-dsl  ──parse──►  AST  ──validate_ast──►  lower  ──
   pre-export pass that rasterizes `.svg` texture slots to PNG and rewrites the
   `TextureRef` paths in a cloned graph, layering the rendered bytes over the caller's
   `TextureSource`. It runs *before* the merge stages in both the GLB and FBX entry points,
-  so the exporter, FBX's raw-path emission, Studio's preview, and `mogen-llm`'s derived
-  PBR maps all see raster only and need no SVG branch of their own. Deliberately uncached
-  and font-free — both so identical sources produce identical bytes.
+  so the exporter and FBX's raw-path emission see raster only and need no SVG branch of
+  their own. `render_svg`/`resolve_svg_size` are also exported standalone so a consumer
+  that never touches a full `SceneGraph` — MoGen Studio's live viewport reads texture
+  paths straight off disk, outside the export pipeline — can rasterize one file at a time
+  instead of re-implementing the renderer. Deliberately uncached and font-free — both so
+  identical sources produce identical bytes.
 - **mogen-llm** — Gemini `generateContent` client (`gemini.rs`), system-instruction assembly
   from grammar + stdlib index + examples (`prompt.rs`), and the repair loop (`repair.rs`)
   that re-feeds JSON diagnostics for up to `max_repair_iters` retries. `embed_seed_header` /
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -974,6 +974,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-url"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "be1e0bca6c3637f992fc1cc7cbc52a78c1ef6db076dbf1059c4323d6a2048376"
+
+[[package]]
 name = "debugid"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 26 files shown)


krazyjakee:
Reviewed this PR (code quality, file size, test coverage, visual correctness) and pushed fixes in 189e5e4.

## Findings, ranked by severity

**1. High — the `mogen` CLI never actually gets SVG support in isolation.** `crates/mogen/Cargo.toml` requested `mogen-export` features `["textures", "merge", "fbx"]` — `textures-svg` was missing, despite the file's own comment stating the CLI wants "every default feature." A `cargo build --release --workspace` (the documented release path) happens to work anyway, because `mogen-export` is itself a workspace root and gets its own default features (which include `textures-svg`) unified in — but that's incidental. `cargo build --release -p mogen` alone, `cargo install --path crates/mogen`, or any packaging step that builds just the CLI silently produces a binary with **no** SVG support: `.svg` textures fail at runtime with "unsupported texture extension." I verified this concretely — an isolated `-p mogen` build before the fix compiled with no `resvg`/`usvg` in the dependency graph; after adding `textures-svg` to the feature list, the same isolated binary builds `examples/features/svg_texture.mog` correctly. Fixed.

**2. High — MoGen Studio's live 3D viewport never rasterized `.svg` either.** Two separate bugs compounded: (a) `mogen-studio/Cargo.toml` had the same missing-feature gap as #1, and (b) even with the feature on, the viewport's texture loader (`viewer/gl_util.rs::try_load_texture`) reads bytes straight off disk and decodes with `image::load_from_memory`, which can't parse SVG at all — it's a completely separate code path from the export pipeline's `svg.rs` pass. The PR's own design table lists "Studio preview" as a consumer the pre-export pass fixes "for free," but Studio's live viewport doesn't call the export pipeline — it builds GL textures directly from `SceneGraph` materials. Net effect: opening the new `examples/features/svg_texture.mog` in Studio would silently render the wall/sign untextured (error only in stderr), while `mogen build` produced a correct GLB. Fixed by exposing `render_svg`/`resolve_svg_size`/`MAX_SVG_SIZE` from `mogen-export` and making the viewport rasterize `.svg` on the fly, threading each material's `texture_size`/`texture_wrap` through `DrawBatch` and widening the texture cache key so two materials sharing one SVG at different sizes don't clobber each other's cached upload.

**3. Medium — `bundle_lods_and_imposter` + `.svg` silently breaks (opt-in feature, not on by default).** The imposter bake goes through `mogen-render`'s own GL texture loader — a third, independent texture-loading implementation with no `TextureSource` hook, so it can neither decode a raw `.svg` nor see the bytes the export pass rasterizes in memory. Combined with an SVG texture, it would try to open the pass's synthetic `*.rasterNNNw.png` path, which never touches disk, surfacing as a confusing "file not found" deep in a stack trace. Full support would mean either duplicating the SVG renderer into `mogen-render` or extracting it into a shared crate `mogen-render` could depend on without creating a cycle (`mogen-export` already depends on `mogen-render` for `ImposterAtlas`) — bigger than warranted for this pass. Added an explicit guard that fails fast with a clear, actionable message instead, a regression test, and a `docs/dsl.md` note calling out the limitation.

**4. Low — test-coverage gap on `MAX_SVG_SIZE`.** The 8192px export cap (`resolve_svg_size`'s upper-bound check) had zero test coverage — only the DSL-level zero/fractional rejection was tested, and DSL lowering doesn't enforce the upper bound (only export does). Added unit tests in `svg.rs` (default, zero, above-max, exactly-at-max) and an integration test asserting the real build fails with a message naming both the offending material and the bound.

## Not flagged (checked and ruled out)
- Duplication of the SVG pre-pass "own-then-shadow" boilerplate between `writer.rs` and `fbx/mod.rs` — this mirrors the codebase's existing convention for the merge stages, not a new anti-pattern.
- FBX raw-path emission — despite the doc comment's "raw path" phrasing, `fbx/texture.rs` embeds actual bytes via the `Video` object's `Content` binary attribute (glTF-analogous), so the rasterized PNG bytes really do get embedded, not just referenced by a synthetic filename that doesn't exist.
- File sizes: nothing introduced or touched crosses 1000 lines (largest changed file is `writer.rs` at 788 lines).

## Visual review
Rendered the PR's example `dots.svg`/`emblem.svg` through the real `render_svg` rasterizer into an imposter atlas (headless GL via Xvfb, 8 yaw angles). Confirmed: the tiled dots pattern joins seamlessly across the `texture_wrap=1` boundary with no visible seam (matches the PR's claimed sub-1/255 residual), and the emblem composites cleanly with correct alpha — no artifacts, stretching, or corruption in any of the 8 baked views.

## Test results
- `cargo test -p mogen-export`: 37 + 5 new = all passing (including the two new regression tests)
- `cargo build --release -p mogen-studio`: clean
- `cargo build --release -p mogen` (isolated, no other `-p`): clean, and the resulting binary correctly builds the SVG example end-to-end

---

Issue #107, opened 2026-08-07
Water should not be a built-in shader — make it use the GLSL mechanism


krazyjakee:
#100 turned the closed `MaterialShader::{Standard,Water}` enum into an open, name-based shader system, and framed the built-in `water` as "just the first client of the same path rather than a special case".

That is the stated design, but it is not what the code does yet. `water` is still a hard-coded branch everywhere that matters, and the generic path it is supposed to exercise is unreachable. Making water a genuine client of the GLSL mechanism would collapse the two paths into one and, in doing so, turn the user-shader feature on.

## Where water is still special-cased

- `crates/mogen-dsl/src/lower/shader.rs:99-102` — seeds a `ShaderDecl` pointing at `stdlib:shaders/water.glsl`, a virtual path with no file behind it. The declaration exists; the GLSL it names does not.
- `crates/mogen-studio/src/viewer/shaders/mesh.rs` — the actual water implementation lives inline in the monolithic fragment program: `water_turbulence()` at :277 and the shading branch around :540-560 and :770-835.
- `crates/mogen-studio/src/viewer/renderer/draw.rs:319` — `shader_id` is literally `if shader_name == WATER { 1 } else { 0 }`. No other shader can ever get an id.
- `crates/mogen-studio/src/viewer/flatten.rs:232` — separate pass keyed on the water name.
- `crates/mogen-validate/src/ast_checks/mod.rs:132` and `crates/mogen-core/src/graph.rs:519` — `builtin_names()` carve-out so `shader="water"` resolves with no declaration.
- `crates/mogen-dsl/src/lower/{cave,terrain}/materials.rs` — generators assign `shader_name = WATER` directly.

## Why this matters beyond tidiness

The injection machinery in `crates/mogen-studio/src/viewer/shaders/user_shader.rs` is written, documented and unit-tested — but **never runs**. `assemble_fs` has exactly one call site, `renderer/mod.rs:240`, and it passes `&[]`. `FIRST_USER_SHADER_ID` is `#[allow(dead_code)]` with `TODO(shader-preview): wired once flatten bakes the used-shader list`.

So today a user can declare a shader, bind it, validate clean, and export correct `node.extras.shader` metadata — and the Studio viewport will render it as plain PBR, silently. `docs/dsl.md:1039` currently tells them otherwise: "MoGen Studio is the one component that actually compiles a declared shader, for live preview."

[...]

PR #108, opened 2026-08-08
svg: rasterization moves below the exporter; the water preset declares its absorption

krazyjakee:
Two changes an engine consuming the lowered `SceneGraph` needs, both of which
also remove a reach-around Studio was already doing.

Requested by a downstream consumer (moggy, krazyjakee/moggy#457), but neither
change is engine-specific — both are about the built-in surfaces being
reachable and self-describing on their own terms.

## 1. `mogen-svg`: the rasterizer moves below the exporter

Rasterization shipped as a pre-export pass (#106). That is still the right
shape for rewriting a `SceneGraph`'s texture paths and layering bytes over a
`TextureSource`, and it is unchanged here. What was in the wrong place was the
renderer underneath it.

Every consumer holding a `Material` and a path sits *beside* the exporter, not
after it. `svg.rs`'s own module docs already named the problem — "the consumer
this pass cannot reach" — and Studio's viewport solved it by reaching back into
`mogen-export` for `is_svg` / `render_svg` / `resolve_svg_size`. An engine that
reads the lowered `SceneGraph` and decodes textures itself cannot do even that
without linking the whole exporter, `oxipng` / `fbxcel` / `meshopt` included,
for three pure functions over `(bytes, size, wrap)`.

So the policy moves into a new `mogen-svg` crate — `mogen-core` plus
`resvg`/`usvg`, both pure Rust — and the pass stays in `mogen-export`, which
re-exports the four public items. `mogen_export::is_svg` and friends still
resolve, so **Studio needs no edit**. `textures-svg` becomes `dep:mogen-svg`.

There is still exactly one implementation of "what pixels does this SVG
produce", which was the point of the original decision; it is simply reachable
now. Verified to cross-compile to `wasm32-unknown-unknown`, as the old module
claimed for itself.

## 2. The built-in `water` preset declares `absorption`

`shader="water"` was a bare on/off switch: the preset was seeded with an empty
param list, so an author could not say *how* absorbing the water is, and a
consumer reading the lowered `Material` found a shader name and nothing else.

It now declares `absorption` (float, default `1.5`), resolved through the
ordinary `ShaderDecl::resolve_param` path — `shader_params (absorption=2.5)`
overrides it exactly as for a user-declared shader. The built-in stays the
first client of the general system rather than acquiring a special case, which
is the property #107 is heading toward anyway.

The default is deliberately not zero: zero is the "behaves exactly like clear
glass" case, so defaulting to it would make `shader="water"` a no-op that
reports success. The test asserts the override wins, that a silent material
gets the default, and — the control — that the default is not the disabled
value.

`docs/dsl.md` documents the parameter in both places water is described.

## Verification

- `cargo test --locked -p mogen-svg -p mogen-export -p mogen-dsl -p mogen-core` — 18 suites, all green (664 in `mogen-dsl`, 27 in `svg_textures`)
- `cargo check --workspace --all-targets` — clean
- `cargo check -p mogen-svg --target wasm32-unknown-unknown` — clean

Not run: the Studio GL paths, which need a display seat this machine may not
have. Studio is untouched by construction — it compiles against the same
re-exported names — but that is an argument, not an observed run.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2763,11 +2763,10 @@ dependencies = [
  "mogen-dsl",
  "mogen-geom",
  "mogen-render",
+ "mogen-svg",
  "oxipng",
- "resvg",
  "serde",
  "serde_json",
- "usvg",
 ]
 
 [[package]]
@@ -2894,6 +2893,17 @@ dependencies = [
  "tiny_http",
  "webbrowser",
  "zip",
+]
+
+[[package]]
+name = "mogen-svg"
+version = "0.1.12"
+dependencies = [
+ "anyhow",
+ "image",
+ "mogen-core",
+ "resvg",
+ "usvg",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ members = [
     "crates/mogen-moghub-client",
     "crates/mogen-registry",
     "crates/mogen-render",
+    "crates/mogen-svg",
     "crates/mogen-studio",
     "crates/mogen-update",
     "crates/mogen-wasm",
@@ -63,6 +64,9 @@ mogen-validate = { path = "crates/mogen-validate", default-features = false }
 mogen-anim = { path = "crates/mogen-anim" }
 mogen-binary = { path = "crates/mogen-binary" }
 mogen-export = { path = "crates/mogen-export", default-features = false }
+# Pure-Rust SVG rasterization. Sits below the exporter so Studio's viewport and
+# downstream engines can resolve the same pixels without linking one.
+mogen-svg = { path = "crates/mogen-svg" }
 mogen-llm = { path = "crates/mogen-llm" }
 mogen-pascal = { path = "crates/mogen-pascal" }
 mogen-moghub-client = { path = "crates/mogen-moghub-client" }
--- a/crates/mogen-core/src/shader.rs
+++ b/crates/mogen-core/src/shader.rs
@@ -28,6 +28,34 @@ pub fn builtin_names() -> &'static [&'static str] {
     &[WATER]
 }
 
+/// Name of the built-in water shader's one parameter: the depth, in world
+/// units, over which transmitted radiance falls to `1/e`.
+pub const WATER_ABSORPTION: &str = "absorption";
+
+/// Default for [`WATER_ABSORPTION`].
+///
+/// Deliberately **not** zero. Zero is the "behaves exactly like clear glass"
+/// case, so defaulting to it would make `shader="water"` a no-op that reports
+/// success. A metre and a half is a plausible pond and is visibly water.
+pub const WATER_ABSORPTION_DEFAULT: f32 = 1.5;
+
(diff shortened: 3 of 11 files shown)


krazyjakee:
Reviewed this PR against code quality, file structure, test coverage, and a visual check of the rendering path it touches. One real gap found and fixed; everything else checks out.

## Findings (ranked by severity)

**1. Low — missing integration test for the new `absorption` param actually reaching the export (fixed)**

The PR's new lowering test (`crates/mogen-dsl/src/lower/tests/shader.rs`) only checks that `ShaderDecl::resolve_param` resolves the right value on the `SceneGraph`. Nothing exercised the actual deliverable this feature is for — per the PR description, "a consumer reading the lowered `Material`" (i.e. a downstream engine) needs the value in the exported glTF's `node.extras.shader.params`, and there was no test pinning that, unlike the existing `declared_shader_params_merge_defaults_with_overrides` coverage for user-declared shaders.

Fixed in `dccebe9`: added `builtin_water_shader_projects_absorption_default_and_override` to `crates/mogen-export/tests/shader.rs`, asserting both the default (`1.5`) and an authored override (`2.5`) land in the exported extras. Also manually verified end-to-end through the real `mogen` CLI (release build, not just `cargo test`):

```json
// node.extras for shader_params (absorption=2.5)
{ "shader": { "name": "water", "params": { "absorption": 2.5 } } }
```

**2. Informational, no action needed — Studio's live preview doesn't consume the new param yet**

`crates/mogen-studio/src/viewer/flatten.rs:586` and `user_shaders.rs:101` still hard-code the built-in water branch to shader id 1 with an empty param list, so `shader_params (absorption=...)` has no visible effect in Studio's GL viewport today. This is a pre-existing, explicitly tracked limitation (`user_shaders.rs` cites krazyjakee/MoGen#107 by name), not a regression introduced by this PR — the data-model-level work here is a reasonable prerequisite for that follow-up. Flagging only so it's not mistaken for a bug if someone tries `shader_params (absorption=…)` in Studio and sees no change.

## Verification performed

- `cargo check --workspace --all-targets --locked` — clean.
- `cargo test --locked -p mogen-svg -p mogen-export -p mogen-dsl -p mogen-core` — all green, including the new test.
- Confirmed the `mogen-export::svg` → `mogen-svg` crate split is behavior-preserving: built `examples/features/svg_texture.mog` through the actual release `mogen` binary and extracted the embedded rasterized textures from the produced GLB. `dots.svg` (wrap/tile mode) and `emblem.svg` (fit mode) both rasterize cleanly with correct colors and seamless wrap-around, no artifacts.
- Confirmed feature wiring (`textures-svg = ["textures", "dep:mogen-svg"]`) and re-exports (`mogen_export::{is_svg, render_svg, resolve_svg_size, MAX_SVG_SIZE}`) are intact — `mogen-studio/src/viewer/gl_util.rs` compiles unchanged against them, matching the PR's "Studio needs no edit" claim.
- No file touched by this PR exceeds 1000 lines; no split needed.

Nice PR — the crate split is clean and the doc comments carry real signal.

---

PR #109, opened 2026-09-05
dsl: fix lowering accuracy and reduce redundant work

krazyjakee:
Imported geometry was resetting LOD at each node, which erased enclosing groups’ `lod=` overrides. Compound physics bodies also replaced explicit group weights with their computed mass. Preserve subtree LOD multipliers across import boundaries, ignore non-finite LOD settings, and retain explicit group weights while still deriving their local centre of gravity.

Reduce lowering overhead by sharing loaded GLB buffers, moving the expanded module registry into its guard, removing child-list clones during physics and shadow traversal, and skipping the physics post-pass when there are no bodies. Keep the physics post-pass in the physics module.

Add six regression tests covering nested imports, sibling isolation, non-finite settings, state restoration after failures, transformed compound bodies, and repeated external meshes. Align the DSL, physics, module, Studio, and API documentation with the implementation. Include runnable import/LOD and transformed compound-weight examples; correct outdated module parameter, import, and physics export rules.

Validation:

- `cargo test -p mogen-dsl`: 687 tests passed on the rebased branch.
- `git diff origin/master --check`: passed.
- `cargo test -p mogen --test goldens`: 14 passed, 8 failed on both this branch and unchanged `master` (`4ab71c2`). All eight failing outputs are byte-identical between the two builds; existing golden files were left unchanged. Failures: `chair`, `chair_array`, `chair_mat`, `chair_module`, `drone`, `mirror_test`, `sword`, `table`.
- `cargo fmt --all -- --check`: reports pre-existing formatting differences, also reproduced on unchanged `master`. New code was formatted without reformatting unrelated files.

Documentation validation:

- `cargo doc -p mogen-dsl --no-deps --document-private-items`: completed; existing private-item and unresolved-link warnings remain in unchanged documentation.
- Built the new Markdown examples with the CLI and inspected the exported GLBs: imported LOD densities match 2.0, 1.0, and 0.25; the compound exports 5 kg and local centroid `[2, 0, 0]`, while its inner group remains 4 kg. An empty physics group retains its explicit weight and omits its centroid.
- Checked all five added documentation links/anchors and confirmed the follow-up Rust edits change comments only.

### DIFF
--- a/crates/mogen-dsl/src/lower.rs
+++ b/crates/mogen-dsl/src/lower.rs
@@ -33,6 +33,7 @@ use anyhow::Result;
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
+use std::rc::Rc;
 
 use mogen_core::{subtree_local_aabb, NodeId, SceneGraph};
 
@@ -47,7 +48,9 @@ use crate::skin_lower::{bind_meshes, lower_skeleton};
 
 use anim::{is_anim_decl, lower_animations};
 use geometry_identity::TessellationCacheGuard;
-use lod::{collect_origin_lods, extract_lod_scale, LodByOriginGuard, LodRequestGuard};
+use lod::{
+    collect_origin_lods, extract_lod_scale, LodByOriginGuard, LodMultiplierGuard, LodRequestGuard,
+};
 use material::collect_materials;
 use node::lower_into;
 use physics::collect_physics;
@@ -76,7 +79,7 @@ thread_local! {
     // in this lowering, keyed by the `src` attribute verbatim. Filled by
     // `collect_mesh_binaries` over the *expanded* AST; read by `primitive_mesh`,
     // which reads the file itself when a spec is absent.
-    pub(super) static MESH_BYTES: RefCell<HashMap<String, Vec<u8>>> =
+    pub(super) static MESH_BYTES: RefCell<HashMap<String, Rc<Vec<u8>>>> =
         RefCell::new(HashMap::new());
 }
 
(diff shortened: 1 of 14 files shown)


krazyjakee:
## Automated review

Reviewed `50145fe` (later rebased onto `68989ba`, the docs-alignment follow-up), focused on code quality, file structure, test coverage, and a visual check since this touches DSL lowering/tessellation.

### Findings

1. **Low — style nit (fixed).** `crates/mogen-dsl/src/lower/primitive.rs` used a fully-qualified `std::rc::Rc::new(...)` inline instead of importing `Rc` at the top of the file, inconsistent with how `lower.rs` imports the same type. Fixed in `5b5e9f4`: added `use std::rc::Rc;` and simplified the call site. No behavior change; full `mogen-dsl` suite re-verified green after the change.

No other issues found:

- **Correctness**: Traced both fixes by hand against the code — the new `LOD_MULTIPLIER` thread-local correctly separates subtree `lod=` accumulation from the per-origin `LOD_SCALE` swap, so an enclosing group's multiplier survives descending into imported geometry (previously erased by the origin guard's `replace`). The compound-physics fix (`b.mass.get_or_insert(total)` instead of an unconditional overwrite) correctly preserves an explicit group `weight=` override, including `weight=0`, while still deriving a real centre of gravity. The `Rc<Vec<u8>>` mesh-byte sharing is read-only past the loader boundary, so aliasing is safe.
- **File size/structure**: all touched/added files are well under 1000 lines (largest is `primitive.rs` at 727); no split needed.
- **Test coverage**: `cargo test -p mogen-dsl` — 687 passed, 0 failed (675 lib + 6 `lod_request.rs` + 6 `mesh_loader.rs`), matching the PR description. The six new regression tests meaningfully cover the described bugs (nested/import LOD isolation, non-finite settings, guard restoration through a failed nested lower, compound weight override + derived COG, mesh-byte dedup across expanded instances).
- **Goldens**: `cargo test -p mogen --test goldens` shows 12 failures in this sandbox vs. the 8 the PR description lists; the extra 4 (`procgen_*`) are Git LFS pointer files that never resolved to real binaries in this environment (unrelated to this change). I confirmed the full failing set — including the 8 real ones (`chair`, `chair_array`, `chair_mat`, `chair_module`, `drone`, `mirror_test`, `sword`, `table`) — reproduces identically byte-for-byte on unmodified `origin/master` via a throwaway worktree, so none of it is a regression from this PR.
- **Clippy**: no new warnings in any file this PR touches; the `approx_constant`/style warnings clippy reports live in unrelated, untouched files (`conform/tests.rs`, `lower/arch/plan.rs`, etc.) and reproduce on `master`.
- **Visual review**: since this changes tessellation density logic, I built `mogen`'s headless `thumbnail` CLI under Xvfb + software GL and rendered `examples/furniture/chair.mog` (renders correctly, no artifacts). I then wrote a targeted repro — an imported module with its own `lod_scale`, wrapped in an enclosing `group (lod=5)` in the importing scene, with a `lod=6` override inside the module itself — and rendered/dumped it on both branches:
  - **This PR**: the imported sphere tessellates at the fully compounded density (49,413 vertices) — smooth.
  - **`origin/master`**: the same sphere collapses to 63 vertices — visibly faceted — because the enclosing group's `lod=5` and the caller's request get discarded when the origin guard resets `LOD_SCALE` on the way into the import.

  This matches the PR's stated bug exactly and confirms the fix works as intended with no visual regression on ordinary (non-imported, non-nested-`lod=`) content like the chair example.

### Conclusion

The change is correct, adequately tested, and does what the description says. Only one trivial style nit was found and fixed; no functional changes were needed.

---

PR #143, opened 2026-09-15
Implement milestone 2: durable modeling sessions and CSG reliability

krazyjakee:
## Problem and behavior

The sports-coupe retry exposed lost review/tool responses, conflicting output contracts, CSG shading and subdivision defects, and a shared refinement workflow that required a custom harness. This PR implements milestone 2 in one branch: typed review recovery, atomic raw-DSL recovery, durable session replay, final CSG normal controls and safe geometric subdivision, scoped external conform bindings, and a supported `mogen session` command plus MCP/Studio integration.

- **#136–#137:** One output contract per phase; strict versioned review schema with lossless findings-array/fence normalization, one judgment-preserving format repair, and raw response retention. Complete DSL tool responses pass through revision-bound Apply with normal scope, lock and validation checks.
- **#138–#139:** Final CSG crease normals (`crease_angle`, default 60°; explicit `faceted`) preserve triangle positions/UV corners. Subdivision uses geometric adjacency separately from UV/render seams, rejects unsupported topology, and checks growth before allocation. The compact sports-body fixture no longer develops extreme spikes at levels 1–2.
- **#140:** Explicit absolute conform target paths cross module instances while ordinary names remain scoped. Missing, inaccessible, ambiguous and cyclic bindings produce actionable diagnostics.
- **#141–#142:** Versioned atomic response journal, per-capture/Apply checkpoints, cumulative budgets, and stage resume shared by Studio and the CLI. MCP returns the CLI's structured artifact report. Outputs distinguish selected, latest and unreviewed candidates. Existing ordinary Draft/auto-refine compatibility is documented.

## Validation

**1,310 targeted tests passed:** 352 LLM, 137 geometry, 680 DSL, 107 validation, 34 CLI/integration/golden tests; one existing CLI test remains ignored. `cargo check -p mogen-studio` passed. Existing unrelated warnings remain.

- Crash injection at 35 durable boundaries verifies real sidecar save/load, saved response/capture reuse and cumulative usage. Additional regressions cover oversized responses, late cancellation, indirect locks and retaining the selected improvement when replay hits a deadline.
- Final-binary scripted CLI smoke ran generation → array review → raw DSL Apply → Finish → improved review in five calls; resume used an empty transcript and made no new calls. MCP stdio schema and structured inspection matched CLI.
- Existing #118 evaluator rendered all seven fixtures. Sports-body tests check levels 0–2, finite unit normals, convex bounds, corner preservation and GLB agreement. Only two golden GLBs change intentionally for CSG normals.

[...]

PR #144, opened 2026-09-15
Fix review findings from #143: swallowed resume errors, self-conform regression, session budget cap

krazyjakee:
## Summary

PR #143 (milestone 2: durable modeling sessions and CSG reliability) merged while an automated code review of it was still in progress. This PR carries the fixes that review found, rebuilt on top of `master` (the `codex/milestone-2-reliability` branch was reset to `master` and these fixes reapplied, since the branch's prior history is already merged).

Full findings list, ranked by severity, is posted as a comment on #143.

## Fixes in this PR

1. **Studio: resume failures reported as success (High).** `run_llm`'s new `resume` branch hardcoded `error: None` in both the `Ok` and `Err` arms of the `resume_session` result, so a failed resume (network error, invalid response, etc.) skipped the error-banner/retry UI entirely and looked like a completed run with the failure text buried in the status line. Now builds a real `LlmErrorInfo` on failure.

2. **conform: self-reference misreported as cyclic (High, regression).** The new topological sort used to order `conform` bindings included each spec in its own blocking-predicate scan, so `conform(target="x", child="x", ...)` could never be selected as runnable and always fell through to the generic "cyclic target/child dependency" bail instead of the specific "cannot be conformed onto itself" diagnostic `apply_conform` already had. Fixed by excluding the self-comparison; added a regression test.

3. **session runner: hardcoded step cap ignored `--calls` budget (Medium).** The tool-correction loop was bounded by a fixed `24` iterations regardless of the session's configured call budget, so a larger `--calls` budget got truncated early and a smaller one let the loop run past what `SessionControl::before_call` already intends to allow. Now derives the bound from the session control's own call limit, falling back to `24` only when there's no control (fixture/mock clients in tests). Added a regression test covering both a smaller and a larger configured budget.

4. **mogen-geom shading: potential out-of-bounds panic (Medium).** `crease_normals` guarded UV copying by length (`has_uvs()`) but guarded `colors`/`joints`/`weights` only by `is_empty()`, so a `Mesh` with a shorter-than-`positions` attribute array would panic. Nothing in `Mesh`'s type enforces that invariant and the function is `pub`. Hardened all three checks to compare lengths; added a regression test.

5. **File size: split `session/tests.rs` (Low/structural).** The file had grown past the project's ~1000-line guideline (1126+ lines). Split into `session/tests/{support,control_budget,locks_scope,dependencies,tools_measure,review_protocol,journal_resume,fixtures_http}.rs`, grouped by concern. Behavior unchanged — same tests, same assertions.

Note: the already-merged commit `d92d3ca` (part of #143) independently added a broader, overlapping `--resume`/`--inspect` sidecar-existence check than the narrower one this review found, so that duplicate fix and its now-redundant test were omitted here.

## Validation

- `cargo test` green across `mogen-llm` (354), `mogen-dsl` (681), `mogen-geom` (138), `mogen-validate` (107), `mogen-studio` (379), and `mogen`'s `session_cli` (5) / `sports_body` (1) suites.
- `cargo fmt --check` clean on every touched file.
- Visual regression check: built the release binary, rendered the `sports_body` fixture at levels 0–2 via Xvfb + software GL (Mesa llvmpipe), and diffed against the checked-in reference PNGs in `benches/quality/reliability/` — negligible pixel differences, no spikes/holes/shading regressions.

## Test plan

- [x] `cargo test -p mogen-llm -p mogen-dsl -p mogen-geom -p mogen-validate -p mogen-studio`
- [x] `cargo test -p mogen --test session_cli --test sports_body`
- [x] `cargo fmt --check` on touched files
- [x] Visual render comparison of `sports_body.mog` at levels 0–2 against checked-in reference images

https://claude.ai/code/session_011Rqk8E9t5SSmE66xgyGv5o

---
_Generated by [Claude Code](https://claude.ai/code/session_011Rqk8E9t5SSmE66xgyGv5o)_


krazyjakee:
## Automated review

Reviewed the diff, ran the full affected test suites, and did a visual render check. Findings ranked by severity below. One gap found and fixed; everything else checks out.

### Findings & fixes

1. **Missing regression test for the highest-severity fix (Low/test-coverage) — fixed.**
   Item 1 in the PR description (`run_llm`'s resume branch hardcoding `error: None` on both `Ok` and `Err`) is the most severe fix in this PR, but unlike items 2–4 it shipped without a test. `run_llm` itself isn't practically unit-testable (it takes a `Sender<LlmMessage>`, a `Credential`, spins up a provider client, etc., and nothing in the crate tests it directly), which likely explains the gap.
   Fix: extracted the `Ok`/`Err` handling into a small `resume_outcome()` helper (pure function over the `resume_session` result, the project, and the checkpoint closure) and added two tests — one confirming a successful resume reports no error, one confirming a failed resume reports a populated, retryable `LlmErrorInfo`, records `stop_reason`, and checkpoints. Pushed as `a690908`.

### Verified correct (no changes needed)

- **`conform` self-reference fix** (`crates/mogen-dsl/src/conform/mod.rs`): the `j != i` exclusion is correct — traced through both the self-target case (single spec, trivially runnable, falls into `apply_conform`'s existing "cannot be conformed onto itself" diagnostic) and a genuine two-spec cycle (`A→B`, `B→A`, still correctly bailed as cyclic). The new test (`conform_rejects_self_target_with_specific_message_not_cyclic`) exercises exactly the regression scenario.
- **Session budget cap fix** (`crates/mogen-llm/src/session/runner.rs`): `max_steps` correctly derives from `SessionControl::limits().calls` (u32), falling back to 24 only when there's no control. The new test drives both a smaller (2) and larger (30) budget than the old constant and confirms the loop stops exactly at the configured limit.
- **`crease_normals` OOB guard** (`crates/mogen-geom/src/shading.rs`): `mesh.has_colors()` and the added `len() == positions.len()` checks for `joints`/`weights` correctly guard the same `i` index used for `positions`/`uvs`. New test constructs a mesh with intentionally short `colors`/`joints`/`weights` arrays and confirms no panic. This is purely defensive for malformed input — all built-in generators only ever emit these arrays empty or full-length, so no behavior change for real content (confirmed visually, see below).
- **`session/tests.rs` split**: diffed function names between the old single file and the 8 new submodules — all 34 original tests plus the 2 shared helpers (`response`, `rev`) carried over intact, only the 1 new regression test added. No hidden loss or duplication.
- **`mogen/src/commands/session.rs`**: change is a no-op variable extraction (`dir.join("final.mog")` → `let entry = ...`), consistent with the pattern already used a few lines above in the non-inspect path.

### Test run

Ran the full suites the PR claims (plus my 2 new tests): `mogen-dsl` 681, `mogen-geom` 138, `mogen-llm` 354, `mogen-validate` 107, `mogen-studio` 381 (379 + 2 new), `mogen`'s `session_cli` 5 / `sports_body` 1 — all green, matching the PR's stated counts.

### Visual review (3D output)

Built the release binary and rendered PNGs via `mogen thumbnail` (headless Mesa software GL) for `sports_body.mog`, `gradient_demo.mog` (vertex-color gradients — the code path closest to the `shading.rs` fix), and `businessman_casual.mog` (rigged character). All three render cleanly: smooth shading, correct gradient colors, no holes/artifacts/black triangles. No regressions observed.

### Note on `cargo fmt --check`

Could not independently verify the PR's `cargo fmt --check` claim — this environment's `rustfmt` (1.8.0, 2026-03-25) reformats large swaths of *unrelated, unmodified* files when run, indicating a version/edition mismatch with whatever produced the repo's existing formatting. Not a finding against this PR; just flagging that I skipped crate-wide `cargo fmt --check` here as unreliable and relied on manual formatting review of the touched lines instead.


---

Issue #173, opened 2026-09-16
Mesh import: Studio import dialog and docs


krazyjakee:
## Problem and intended behavior

Studio users should be able to open or drag in OBJ/glTF/FBX files and get editable `.mog`, with a preview of how faithful the conversion is.

## Scope

- File → Import (and drag-and-drop) for mesh formats. Offer options for tolerance, fitting on/off, up axis and scale.
- Show the import report: passes applied, fallbacks to `poly`, per-node error, and skipped features.
- Update `docs/cli.md` and `docs/studio.md`, keeping the heading anchors in `mogen-studio/src/docs.rs` in sync.

## Acceptance criteria

- [ ] Import runs off the UI thread and opens the result in a new tab.
- [ ] Doc anchor links resolve.

Issue #171, opened 2026-09-16
Mesh import: simple CSG recovery (research)


krazyjakee:
## Problem and intended behavior

Some shapes are really boolean operations: a plate with drilled holes, a hollowed box, a window cut-out. Recovering `difference`/`intersect` makes the source express the author's intent. This is research-grade, so the scope is limited to simple cases.

## Scope

- After primitive detection, look for concave regions that match a subtracted primitive (e.g. a cylindrical bore through a box) and propose `difference(base, cutter)`.
- Check each candidate by lowering it through `mogen-geom` CSG and running the fidelity gate. Pick the tree with the fewest nodes that passes.
- Look into InverseCSG and CSGNet approaches for later expansion. Write up findings on this issue.

## Acceptance criteria

- [ ] A box with through-holes and a hollowed box are recovered as `difference` trees within tolerance.
- [ ] If no candidate passes, the previous passes' output is left unchanged.

Issue #168, opened 2026-09-16
Mesh import: whole-component primitive fitting (box, sphere, cylinder, cone, torus, plane)


krazyjakee:
## Problem and intended behavior

Many game and CAD-style assets are built from separate parts that are basic primitives. Detecting these gives compact, readable, LLM-editable `.mog` instead of long lists of numbers.

## Scope

- Split meshes into connected components.
- For each component, try fitting box (OBB with 6 planar face clusters), sphere, cylinder/capsule, cone, torus and plane/quad, using least squares with an analytic initial guess.
- Recover segment/ring counts where the source is clearly tessellated (e.g. a 16-sided cylinder), so rendering stays faithful.
- Accept a fit only if it passes the fidelity gate. Otherwise keep `poly`. Output `pos`/`rot`/size attributes using MoGen conventions.
- Add a flag to turn fitting off (`--no-fit`) and per-primitive allow lists.

## Acceptance criteria

- [ ] Primitives built by MoGen, re-imported from GLB, come back as the same primitive kinds with matching parameters (within tolerance).
- [ ] Noisy or bevelled shapes fall back to `poly` instead of being forced into a primitive.
- [ ] Results are deterministic.

Issue #167, opened 2026-09-16
Mesh import: geometric fidelity metric and tolerance gate


krazyjakee:
## Problem and intended behavior

Every conversion pass that replaces triangles with a primitive needs an objective accept/reject check. Otherwise recovered shapes can drift away from the source without anyone noticing.

## Scope

- Compute a symmetric Hausdorff distance (plus mean or RMS surface distance) between source triangles and lowered `.mog` geometry, sampling points deterministically. Scale the tolerance relative to the bounding-box diagonal.
- `--tolerance <rel|abs>` CLI flag. Passes use it to accept a fit, or fall back to `poly`.
- Add a `--report` JSON output with the error for each node and the pass that produced it.
- A test harness that imports, lowers and measures, used by the golden tests.

## Acceptance criteria

- [ ] The metric is deterministic, and fast enough to run for every candidate fit on assets of 100k triangles or more.
- [ ] The reported errors for all golden fixtures are within tolerance.

PR #175, opened 2026-09-18
Mesh import: mogen-import crate + glTF/GLB reader

### LINKED ISSUE #162: Mesh import: mogen-import crate, shared mesh IR, and CLI dispatch
## Problem and intended behavior

`mogen import` only handles pascalorg/editor JSON. Existing OBJ, glTF/GLB and FBX assets can't be brought into MoGen. We need a single entry point that reads a mesh file into a format-neutral intermediate form and emits `.mog` source, so format readers and conversion passes can be built separately.

## Scope

- New `mogen-import` crate (depends on `mogen-core`/`mogen-geom` only, and emits `.mog` source text, like `mogen-pascal`).
- Mesh IR: node tree with local transforms, per-node triangle meshes (positions, indices, normals, UVs), material table, and source names.
- `mogen import <file>` picks the reader from the file extension (`.json` → pascal, `.glb`/`.gltf`/`.obj`/`.fbx` → mesh import). Existing flags stay the same.
- A header comment in the output lists everything skipped, dropped or approximated, the same way the pascal importer does it.
- A pass pipeline (`ingest → normalize → convert → emit`) so primitive recovery passes can be added later.

## Acceptance criteria

- [ ] `mogen import foo.glb` produces a `.mog` that builds without errors (using the reader from the glTF issue).
- [ ] The pascal import behaviour and its tests are unchanged.
- [ ] Output is byte-identical across runs (no HashMap iteration order).
- [ ] `docs/cli.md` `import` section is updated and the Studio docs anchors still resolve.

## Related work

Precedent: `crates/mogen-pascal`, `docs/cli.md#import`.

### LINKED ISSUE #163: Mesh import: glTF / GLB reader
## Problem and intended behavior

glTF is the easiest format to support first (the `gltf` crate is already a workspace dependency), and it's the format MoGen itself exports, which makes round-trip testing possible.

## Scope

- Read `.glb` and `.gltf` (with external buffers) into the mesh IR: node hierarchy, TRS or matrix transforms, all primitives with mode TRIANGLES (report and skip other modes), POSITION/NORMAL/TEXCOORD_0, indexed and non-indexed.
- Materials: base colour factor, metallic/roughness, emissive, alpha mode, double-sided. Extract base colour textures next to the output and reference them.
- Cameras, lights (KHR_lights_punctual), skins and animations are reported as skipped for now (see the skins/animation issue).
- Pick up names for nodes, meshes and materials, sanitised into valid, unique `.mog` identifiers in a deterministic way.

## Acceptance criteria

- [ ] Every GLB in the golden set imports, and the resulting `.mog` builds.
- [ ] Round-trip: MoGen GLB → import → build stays within the fidelity tolerance.
- [ ] Unsupported features show up in the header report instead of failing silently.

krazyjakee:
Closes #162, closes #163. First step of the mesh-import track (#162–#173).

## What

`mogen import model.glb` (or `.gltf`) now writes editable `.mog` source. `.json` still goes to the pascal importer, which is unchanged. `.obj` / `.fbx` are recognised and get a clear "not supported yet" message (#164, #165).

New crate `crates/mogen-import`. It emits text only and does not link the compiler.

| stage | module | what it does |
|---|---|---|
| ingest | `gltf_reader.rs` | reads the default scene into `ir::MeshScene`: node tree, TRS (matrices decomposed), per-material triangle primitives, material table. Strips and fans are converted to triangle lists. PNG/JPEG textures (buffer view, data URI or external file) are copied out byte-for-byte |
| normalize | `normalize.rs` | names become `[A-Za-z0-9_]`, made unique with `_2`, `_3`, … in source order |
| convert | `convert.rs` | `Pass` trait plus an ordered pass list. Current passes: drop degenerate / out-of-range triangles, compact vertices, prune empty subtrees and unused meshes/materials. Primitive fitting, symmetry and CSG recovery (#168–#171) will be added here as more passes |
| emit | `emit.rs` | a `group` per node, a `poly` per primitive (`uvs=` from TEXCOORD_0), and a parameterless `module` + `use` for meshes placed by more than one node. Materials get every factor written out explicitly |

The header comment lists everything that was skipped or approximated: cameras, lights, skins, animations, morph targets, vertex colours, extra UV sets, unsupported extensions, nodes without geometry, and normal/occlusion scale. The CLI prints the same list and writes the textures to `<name>_textures/` beside the output.

For MoGen's own GLBs, the `role` and `tags` stored in node `extras` are carried back. That keeps `floating` exemptions working. Debug POI markers, which the validator exempts by `kind`, are tagged `floating` so they stay exempt.

## Tests

- `tests/round_trip.rs`: every committed GLB (23 files across `examples/` and the CLI fixtures) is imported, validated, lowered and rebuilt with no errors. The rebuilt file has the same triangle count and the same world bounds (within 0.1% of the model's size). Importing twice gives byte-identical text.
- `tests/gltf_features.rs`: a hand-built glTF covering an embedded PNG, a triangle strip, non-indexed geometry, a points primitive, a matrix transform, a camera, `KHR_texture_transform`, name collisions and a shared mesh. It checks the exact emitted materials and structure plus the header report, then builds the result and confirms the texture is in the rebuilt GLB.
- Unit tests for number formatting (the grammar has no exponent syntax), name sanitising, the skip report, strip/fan conversion, Euler round-trip and extras parsing.
- The `mogen-pascal`, `mogen` (goldens included) and Studio docs tests all pass unchanged. Clippy is clean on the new crate.

Visual check with `mogen thumbnail` (original build next to imported → rebuilt): sword (hard edges), drone (hierarchy and shared meshes) and metaballs (smooth shading) look the same.

## Not in this PR (tracked)

- Welding / crease-angle splitting and configurable precision in the CLI: #166
- A real fidelity metric in place of the bounds check: #167
- A Studio import dialog: #173. Studio's File menu still offers only the Pascal import.

Docs: the `docs/cli.md#import` section is rewritten (the anchor is unchanged), and there are new README and CLAUDE.md entries.


krazyjakee:
Reviewed this PR (code quality, file size, test coverage, and a visual check of the import round-trip). Ran `cargo build --workspace`, `cargo test -p mogen-import`, and `cargo clippy -p mogen-import --all-targets` — all clean.

## Findings, ranked by severity

**1. [Fixed, pushed] `round_trip.rs` crashed on unresolved Git LFS pointer files — medium severity, test robustness.**
`golden_glbs()` walks every `.glb` under `examples/` and `crates/mogen/tests/fixtures/`, but not all of those are mirrored in plain git the way the compiler's own goldens are (see the comment in `.gitattributes` explaining why the goldens fixtures are kept out of LFS). In a checkout that hasn't run `git lfs pull` — this sandbox, and presumably any fresh CI or contributor clone — `examples/nature/terrain.glb` is just an LFS pointer stub, and the test panicked with a confusing `not a valid glTF / GLB file: expected value at line 1 column 1` instead of a clear error. Fixed by detecting LFS pointer files and skipping them with an explicit message (`crates/mogen-import/tests/round_trip.rs`, commit bd21888), matching the "works without `git lfs pull`" guarantee the rest of the suite relies on. `cargo test -p mogen-import` is green after the fix (14 + 2 + 1 + 1 doctest, all passing).

**2. External image URIs are resolved with a plain `base.join(...)` — low severity, worth a look.**
`gltf_reader.rs::image_bytes` does `base.join(percent_decode(uri))` for external (non-data-URI) images with no containment check. A crafted `.gltf` with an image URI like `../../../etc/foo.png` (or an absolute path, which replaces `base` entirely under `Path::join`) would have the importer attempt to read outside the input's directory. Actual exfiltration risk is low — the bytes still have to sniff as PNG/JPEG to end up written to the output textures folder — but since this pipeline may process externally-sourced GLBs (not just self-authored ones), it's worth at least a code comment noting the trust boundary, or a check that the resolved path stays under `base`. Not blocking; didn't fix this myself since the right answer (hard error vs. silent clamp) is a product decision.

**3. Sampler wrap/filter settings are dropped without a report note — low severity, minor completeness gap.**
Every other lossy decision (normal scale, occlusion strength, extra UV sets, vertex colours, non-TEXCOORD_0 textures) gets a `report.note(...)` so it shows up in the header comment. Texture sampler settings (wrap mode, filtering) have no equivalent — they're silently ignored. Minor, and arguably fine as an implicit "textures are just images" simplification, but inconsistent with how carefully everything else here is reported.

## Visual review

Imported `examples/vehicles/drone.glb` (deep hierarchy + shared meshes) and `examples/weapons/sword.glb` (hard edges) back through the pipeline (`import` → `thumbnail`, rendered headless via Xvfb) and diffed against thumbnails of the original hand-authored `.mog` sources. Both pairs are visually identical — geometry, hierarchy, materials, and hard/soft edge shading all survive the round trip with no artifacts.

## Everything else

File sizes are all well under the 1000-line split threshold (largest is `gltf_reader.rs` at 573 lines). The pipeline (ingest → normalize → convert → emit) is cleanly separated, deterministic (no map iteration, explicit `Vec`-based namers), and the pass-based `convert.rs` design leaves clean room for the recovery passes tracked in follow-up issues. Test coverage is strong: unit tests per module plus two solid integration tests (23 real GLBs round-tripped, plus a hand-built glTF exercising strips/fans/points/matrices/textures/shared-mesh/extension-skip paths). No other correctness issues found.


---

